PR #5, opened 2026-05-23
Update extract.py

miller73mb-sys:
Migliorata la gestione dell'estrazione testo con PyPDF2.

### DIFF
--- a/scripts/extract.py
+++ b/scripts/extract.py
@@ -374,23 +374,21 @@ def extract_with_pdftotext(pdf_path: str) -> str | None:
     return None
 
 
-def extract_with_pypdf2(pdf_path: str) -> str | None:
-    try:
+def extract_with_pypdf2(pdf_path: str) -> str | None:  
+try:
         import PyPDF2
         text_parts = []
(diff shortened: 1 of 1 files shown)


virgiliojr94:
Thanks for the interest in improving the PDF extraction path. I can't merge this one, though — the change moves `try:` to column 0, outside the function body, which makes `scripts/extract.py` fail to parse:

```
$ python3 -m py_compile scripts/extract.py
IndentationError: expected an indented block after function definition on line 377
```

That breaks the entire script for every format, not just PDF. Aside from the indentation, the only functional change is the added comment — `page.extract_text() or "" ` was already handling per-page extraction — and it drops the explicit `except ImportError: return None` branch.

Separately, `extract.py` is being refactored into a modular `extractor/` package in #8, so this monolithic file is on its way out.

Closing for those reasons. If you'd like to contribute a real improvement to the PyPDF2 path, a good follow-up would be page-level error logging or a fallback when `extract_text()` returns empty — happy to review that. Thanks again.

PR #12, opened 2026-06-03
fix: resolve OPF-relative hrefs in EPUB zipfile fallback

### LINKED ISSUE #11: EPUB stdlib fallback fails when OPF is not at archive root (OEBPS-relative hrefs)
## Summary

The dependency-free EPUB fallback `extract_with_zipfile` (`scripts/extractor/parsers/epub.py`) returns no text for EPUBs whose OPF package document is not at the archive root — i.e. content under `OEBPS/` or `OPS/`, which is how LibreOffice, Calibre, and most publisher EPUBs are structured. The ebooklib path is unaffected; this only impacts the no-dependency fallback.

This is **pre-existing** (the same limitation is present in the pre-refactor `master` extractor) — filing as a follow-up to #8, not a regression in it.

## Root cause

`extract_with_zipfile` reads the OPF and extracts spine hrefs with:

```python
spine_order = re.findall(r'href=["\']([^"\']+\.(?:xhtml|html))["\']', opf_text)
```

OPF hrefs are **relative to the OPF file's own directory**, not the archive root. For an OPF at `OEBPS/content.opf` with `href="sections/section0001.xhtml"`, the actual zip entry is `OEBPS/sections/section0001.xhtml`. The code then calls `zf.read("sections/section0001.xhtml")`, which raises `KeyError`, is swallowed by the inner `except: continue`, leaves `parts` empty, and returns `None`.

## Reproduction

```bash
# Generate a real EPUB (OPF lands under OEBPS/)
soffice --headless --convert-to odt  --outdir /tmp /tmp/book.html
soffice --headless --convert-to epub --outdir /tmp /tmp/book.odt

python3 -c "import sys; sys.path.insert(0,'scripts'); \
from extractor.parsers.epub import extract_with_zipfile as e; \
print(e('/tmp/book.epub'))"
# -> None  (expected: the chapter text)
```

Zip layout of such an EPUB:
```
mimetype
META-INF/container.xml
OEBPS/content.opf
OEBPS/toc.xhtml
OEBPS/sections/section0001.xhtml
```
Regex matches `sections/section0001.xhtml`; the real entry is `OEBPS/sections/section0001.xhtml`.

## Suggested fix

Resolve each href against the OPF's directory before reading:

```python
import posixpath
opf_name = opf_files[0]
opf_dir = posixpath.dirname(opf_name)  # e.g. "OEBPS"
...
for href in spine_order:
    name = posixpath.normpath(posixpath.join(opf_dir, href)) if opf_dir else href
    raw = zf.read(name).decode("utf-8", errors="replace")
```

Also worth: read `META-INF/container.xml` to locate the OPF (instead of globbing `*.opf`), since that's the spec-defined entry point.

## Test gap

`tests/test_extractor.py::test_epub_extraction_via_zipfile_fallback` uses a minimal EPUB with a root-level OPF, so it doesn't exercise the `OEBPS/`-relative case. A fixture with the OPF in a subdirectory would catch this.

Marcelluxx:
## Closes
Closes #11

---

## Description
The dependency-free EPUB fallback (extract_with_zipfile) returned no text for EPUBs whose OPF package document lives in a subdirectory (e.g. OEBPS/content.opf), which is the standard layout produced by LibreOffice, Calibre, and most publishers.

### Root cause
OPF manifest href attributes are **relative to the OPF file's own directory**, not the archive root. The code was reading them as-is against the zip, causing KeyError on zf.read(), which was silently swallowed by the inner except: continue — leaving parts empty and returning None.

### Changes

**scripts/extractor/parsers/epub.py:**
- Added _find_opf_path() helper that reads META-INF/container.xml (the EPUB spec-defined entry point) to locate the OPF, falling back to a *.opf glob scan
- Resolve each manifest href against the OPF directory using posixpath.join/
ormpath before reading from the zip
- Applied the same OPF lookup in count_epub_chapters() for consistency

**	ests/test_extractor.py:**
- Added _make_oebps_epub() fixture generating an EPUB with OEBPS/ subdirectory layout, container.xml, and nested sections/ chapter files
- Added 5 new tests in TestEpubOpfRelativePaths:
  - Zipfile fallback resolves OEBPS-relative paths correctly
  - Full end-to-end extraction succeeds with OEBPS layout
  - _find_opf_path prefers container.xml over glob
  - count_epub_chapters works with subdirectory OPF
  - Root-level OPF regression check (no breakage)

### Verification
All **38 tests** pass (33 existing + 5 new).

### DIFF
--- a/scripts/extractor/parsers/epub.py
+++ b/scripts/extractor/parsers/epub.py
@@ -1,3 +1,4 @@
+import posixpath
 import re
 import zipfile
 from extractor.parsers.html import _HTMLTextExtractor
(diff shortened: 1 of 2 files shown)


virgiliojr94 (approved):
Verified against the exact failing case from #11. I re-ran my original repro — a real LibreOffice-generated EPUB with an `OEBPS/content.opf` layout and nested `OEBPS/sections/` chapter files:

- `extract_with_zipfile` now returns 882 chars (was `None`) with all three chapters present.
- End-to-end `extract.py book.epub --install-missing no` → exit 0, 139 words, correct content (Backpressure / Retries / Bulkheads).
- Full suite: **38/38** pass.

The fix is spec-correct: `_find_opf_path` reads `META-INF/container.xml` (the defined entry point) with a `.opf` glob fallback, and hrefs are resolved against the OPF directory via `posixpath.join`/`normpath`. Same lookup applied to `count_epub_chapters` for consistency, and the root-level-OPF regression case is covered.

Nit (non-blocking, no need to hold the merge): `except (KeyError, Exception):` in `_find_opf_path` is redundant — `Exception` already covers `KeyError`. Worth simplifying to `except Exception:` next time you touch it.

Approving and merging. Thanks for the quick, well-tested turnaround — this closes the loop from the #8 review.

PR #16, opened 2026-06-04
test: tripwire — reintroduce global sort to verify CI catches it

virgiliojr94:
Deliberate regression to validate the CI test gate. Re-adds the global alphabetical sort in resolve_input_files that #8 removed, breaking explicit input-order preservation. Expect the test job to fail on TestInputOrderPreservation. To be reverted / closed without merge.


virgiliojr94:
Tripwire confirmed: the test job failed across py3.10–3.13 on the deliberate input-order regression, while lint/smoke/validate-skill stayed green. CI gate works. Closing without merge as planned.

PR #24, opened 2026-06-08
feat(skill): emit machine-readable index.json for one-hop topic lookup

virgiliojr94:
Step 9.5 generates index.json (topic->chapter file) so the agent resolves a topic in one small read instead of scanning prose — attacks the ~9.8K-token ToC discovery cost measured by discovery_tax.py. How-to-Use/fold-in updated to consult+sync it. tools/validate_index.py + tests catch drift.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Running `/book-to-skill your-book.pdf` (or a folder, glob, or list of files) cre
 | File | Purpose | Size |
 |------|---------|------|
 | `SKILL.md` | Core mental models + chapter index | ~4,000 tokens |
+| `index.json` | Machine-readable topic→chapter map for one-hop lookup | tiny |
 | `chapters/ch01-*.md` … | One file per chapter, loaded on-demand | ~1,000 tokens each |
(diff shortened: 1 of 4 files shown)


virgiliojr94:
On hold per empirical audit: spec-only (no runtime proof the host agent reads index.json) and adds to the always-loaded SKILL.md budget. Reopen for merge once there is measured runtime benefit.

virgiliojr94:
Closing — measured, not merging. A true runtime test shows index.json gives **no token gain, it's a net loss**:

| same index (10 ch, 25 topics) | tokens |
|---|---|
| Markdown Topic Index (already in SKILL.md) | 250 |
| index.json (minified) | 506 |
| index.json (pretty, as specced) | 926 |

JSON overhead (keys, quotes, indentation) makes it 2–3.7× heavier than the Markdown index it would replace — which is already resident in SKILL.md, so resolution is effectively free today. The ~9.8K ToC cost that motivated this was the **raw PDF** ToC, already solved by having a compiled SKILL.md at all. Only non-token benefit would be machine-parse reliability, which isn't worth the weight. Markdown index stays.

PR #25, opened 2026-06-08
feat(skill): record per-chapter provenance for fact-checking

virgiliojr94:
Each chapter ends with a provenance comment (source + full_text.txt line range) so any claim is traceable to the extracted text — answers the 'degraded signal you can't verify' critique. Honest caveat: ranges are per-generation, shift on re-extraction.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -392,7 +392,20 @@ Create `$SKILLS_HOME/<skill_name>/chapters/ch<NN>-<slug>.md` using the structure
 ## Connects To
 - **Ch N**: <why this chapter relates>
 - **<Concept>**: <external concept or standard it connects with>
+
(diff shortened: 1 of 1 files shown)


virgiliojr94:
On hold per empirical audit: provenance is spec-only and not enforced; line offsets go stale on re-extraction. Reopen once a validation exercises it.

virgiliojr94:
Closing — measured, not merging. Provenance is by design **pure cost** (~20–40 tokens per chapter) for an audit benefit, with no token gain and no enforcement; offsets also go stale on re-extraction. Not worth adding to every generated chapter without a concrete fact-checking workflow that uses it. Can revisit if such a workflow lands.

PR #30, opened 2026-06-11
add support to github copilot

ShawnXxy:
## What & why
Makes `book-to-skill` a tri-agent project. Previously the converter targeted Claude Code (primary) with Amp also supported (see CHANGELOG #13/#14). This PR adds **GitHub Copilot CLI** as a first-class target so the same `SKILL.md` discovers, installs, and runs identically across all three hosts via the open [Agent Skills](https://github.com/agentskills/agentskills) standard.

No Python source under `scripts/` was touched — the extractor is host-agnostic. All changes are in the SKILL.md generator spec, the validator, the README, and the architecture doc.

### `SKILL.md`
- **Frontmatter** trimmed to the open-standard minimum (`name` + `description`). The previous `compatibility`, `allowed-tools`, and `argument-hint` keys were a mix of Claude- and Amp-flavored tokens that Copilot silently ignored; moved to an HTML comment for human readers. Both Claude and Copilot interpret missing `allowed-tools` as "no restriction" and prompt on first use of shell/Read/Write.
- **Description** updated to mention all three hosts so each agent's auto-loader picks it up.
- **Skill Locations** section now lists 8 discovery paths in Copilot-first order: `~/.copilot/skills/`, `~/.agents/skills/`, `~/.claude/skills/`, then the four project-local roots, then the two Amp-specific roots.
- **Step 2** (`SCRIPT_PATH` resolution): extended the bash probe ladder to walk all 8 candidates.
- **Step 5** (where to write the generated skill): replaced the Claude-only default with a per-host selection table. Rules: single existing root → use it; none → ask once per session; project-local present → use it; unknown host → ask which agent.
- **Step 9** (generated child-skill template): dropped `allowed-tools: [Read, Grep]` and `argument-hint` from its frontmatter (same rationale as parent).
- **Step 10** (cleanup report): added a per-host reload hint + an opt-in `gh skill publish` line.

### `tools/validate_skill.py`
- Added a `--lens claude|copilot|amp` flag. Each lens is a table of: recognized tool tokens, recognized frontmatter keys, bash-equivalent tool names, reserved name words, and unknown-tool severity.
- Behavior:
  - `claude` — same rules as before (back-compat: this is the default, so existing CI keeps passing without changes).
  - `copilot` — recognizes `shell`/`bash`/`write`; unknown tokens are warned, not errored (Copilot treats them as MCP server names).
  - `amp` — Claude's tool set plus `shell_command`; also accepts `compatibility`/`argument-hint` keys that other hosts ignore.
- Each lens exits non-zero only on rules that would actually break the skill on that host.

### `README.md`
- Headline updated to call out GitHub Copilot CLI, Amp, and Claude Code.
- Badge changed from "Claude_Code-Skill" to "Agent_Skills-Open_Standard".
- "What it generates" and "How it works" diagram now show all three install destinations.
- Install section: new Copilot CLI block first, then the cross-agent `~/.agents/skills/` path, then the existing Claude Code block (preserved unchanged for current users).
- Usage section: agent-neutral wording + `/skills reload` note for Copilot.

### `docs/ARCHITECTURE.md`
- Output diagram shows the per-host destination paths.
- Key-components table notes the new `--lens` flag on the validator.

## Checklist
- [ ] One focused change
- [ ] Tests added/updated for behavior changes
- [ ] `ruff check .` clean
- [ ] `pytest -q` green
- [ ] `python3 tools/validate_skill.py SKILL.md` passes (if SKILL.md changed)
- [ ] `CHANGELOG.md` updated under `## [Unreleased]`
- [ ] No raw book text shipped; no net SKILL.md bloat without justification

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 <h1 align="center">📚 book-to-skill</h1>
 
 <p align="center">
-  <strong>Turn any technical book, document folder, or collection of sources into a unified Claude Code skill — ready to study, reference, and use while you work.</strong>
+  <strong>Turn any technical book, document folder, or collection of sources into a unified agent skill — ready to study, reference, and use while you work in GitHub Copilot CLI, Amp, or Claude Code.</strong>
 </p>
 
 <p align="center">
   <a href="https://github.com/virgiliojr94/book-to-skill/releases"><img src="https://img.shields.io/github/v/release/virgiliojr94/book-to-skill?style=for-the-badge&color=blueviolet" alt="Latest release"></a>
-  <img src="https://img.shields.io/badge/Claude_Code-Skill-blueviolet?style=for-the-badge" alt="Claude Code Skill">
+  <img src="https://img.shields.io/badge/Agent_Skills-Open_Standard-blueviolet?style=for-the-badge" alt="Agent Skills standard">
   <img src="https://img.shields.io/badge/PDF%20%E2%80%A2%20EPUB%20%E2%80%A2%20DOCX%20%E2%80%A2%20MD%20%E2%80%A2%20HTML%20%E2%80%A2%20RTF%20%E2%80%A2%20MOBI-supported-green?style=for-the-badge" alt="Formats supported">
   <img src="https://img.shields.io/badge/License-MIT-blue?style=for-the-badge" alt="MIT License">
 </p>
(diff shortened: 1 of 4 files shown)


virgiliojr94 (changes requested):
Great addition — the tri-agent support (8 discovery paths, Step 5 per-host table, and the `--lens` validator with `claude` as the back-compat default) is clean and CI stays green. Happy to merge once one thing is addressed.

**Please keep `allowed-tools` in the frontmatter** (parent `SKILL.md` and the Step 9 child template), using Claude-native tokens only:

```yaml
allowed-tools:
  - Bash
  - Read
  - Write
  - Glob
  - Grep
```

Reasoning: #18 deliberately restored the Bash grant so Claude Code pre-authorizes shell/file tools and doesn't prompt on every `extract.py` run. Removing the key reintroduces a permission prompt on first use of each tool for Claude users (the primary audience). And per your own note, Copilot *ignores* unknown frontmatter keys — so keeping `allowed-tools` costs Copilot nothing while preserving the Claude UX. The only real wart in the old value was the Amp-flavored `shell_command` token (which the Claude lens warns on); dropping just that token fixes the warning without losing the grant.

Everything else looks good to ship — the multi-host work is a genuinely valuable direction. Thank you for the careful PR description, it made review fast.

virgiliojr94 (approved):
Ran all three validator lenses on your SKILL.md locally — `claude`, `copilot`, and `amp` all pass (only the pre-existing body>500 soft warning). So the change is conformant on every host; my earlier note about `allowed-tools` was a UX preference (preserving the Claude Bash pre-grant from #18), not a conformance blocker.

Given it's green everywhere and the multi-host support is valuable, merging as-is. If Claude users report permission-prompt friction from the missing `allowed-tools`, we'll restore it as a fast-follow (Claude-native tokens only; Copilot ignores the key anyway). Thanks for the clean, well-documented PR.

PR #34, opened 2026-06-12
fix(extractor): defer annotation evaluation so import works on Python 3.9

RandMelville:
PR 1 of 2 — runtime robustness.

## What
Adds `from __future__ import annotations` to every module that uses PEP 604 unions (`str | None`) in signatures: `scripts/extractor/{utils,dependencies}.py`, `scripts/extractor/parsers/*.py`, and `tools/discovery_tax.py`. Also adds `"3.9"` to the CI test matrix (per review) so the fix is guarded, not just promised.

## Why (the learning)
Function-signature annotations are evaluated **at import time**, not lazily. `str | None` is only a valid expression from Python 3.10 on. On 3.9 it raises `TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'` **at import** — and since the package fails to import, the whole suite fails during pytest collection.

The current CI matrix (3.10–3.13) never exercised 3.9, so nothing caught it. `from __future__ import annotations` makes all annotations lazy strings (PEP 563): zero runtime cost and the `|` is never evaluated.

## Tradeoff
Minimal diff (one line per file), no behavior change on 3.10+. Positive side effect: the package now imports and runs on 3.9 too — which is why PR 2 declares `requires-python = ">=3.9"`.

## Evidence
`pytest` on **Python 3.9.6**: before, collection failed with the `TypeError`; after, **60 passed**. `ruff --select E9,F` clean. The new 3.9 matrix entry locks this in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to **book-to-skill** are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **Package now imports on interpreters that evaluate annotations eagerly.**
+  Added `from __future__ import annotations` to every module using PEP 604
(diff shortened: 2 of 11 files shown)


virgiliojr94 (commented):
Let's go with this approach — supporting 3.9 via deferred annotations is clean, and I verified there's no other 3.10+ runtime usage (no `match`, no isinstance-unions, no runtime union values), so it genuinely imports and runs on 3.9.

One thing before merge, to make the support real instead of a promise: the CI matrix only runs 3.10–3.13, so nothing guards 3.9. Could you add `"3.9"` to the matrix in `.github/workflows/ci.yml`? It's the one-line change that keeps this from silently re-breaking the first time someone uses a 3.10+ feature outside an annotation. With 3.9 green in CI, this is good to merge. Thanks for the fix and the clear write-up!

RandMelville:
Done — added `"3.9"` to the CI matrix in `.github/workflows/ci.yml` (commit 9d1b9ad), so the support is guarded by CI rather than just promised. The 3.9 job runs the full suite (60 tests) on the deferred-annotations fix.

Folded it into this PR instead of a separate one, as you suggested, so the fix and its guard land as one coherent unit. Thanks for the quick review!

PR #35, opened 2026-06-12
chore: add pyproject.toml (requires-python + ruff/pytest config)

RandMelville:
PR 2 of 2 — explicit runtime contract.

## What
Adds a minimal `pyproject.toml`: project metadata, `requires-python = ">=3.9"`, and the `ruff`/`pytest` config in one place.

## Why (the learning)
Today versions and rules live scattered (flags in `ci.yml`, support only implicit). Centralizing buys two things:

1. **Runtime contract**: `requires-python` makes `pip` refuse an incompatible interpreter with a clear message, instead of the cryptic import-time `TypeError`. (`>=3.9` because the package runs on 3.9 — see PR 1.)
2. **Dev == CI**: with `[tool.ruff]` and `[tool.pytest.ini_options]`, running `ruff check .` and `pytest` locally uses exactly what CI uses, no flags to memorize.

## Tradeoff
**Metadata only** — no build backend, no `console_scripts`. The tool stays CLI-first (`python3 scripts/extract.py`); packaging to PyPI is not worth the complexity now. Keeps the surface minimal.

## Evidence
`ruff check .` reads the config and passes with no flags. `pytest` discovers tests via `testpaths`. (The suite going green on 3.9 lands with PR 1.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **`pyproject.toml`** declaring `requires-python = ">=3.9"` and centralizing the
+  ruff/pytest config so `ruff check .` and `pytest` locally match CI without
+  passing flags by hand. Metadata only — no build backend or console scripts; the
+  tool stays CLI-first via `python3 scripts/extract.py`.
 - Structure detection now recognizes Markdown/AsciiDoc ATX headings (`#`, `==`)
   as chapters when no numeric "Chapter N" headings are present, fixing a
   zero-chapter result for `.md`/`.adoc` sources. Headings inside fenced code
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,18 @@
+[project]
+name = "book-to-skill"
+version = "1.1.0"
+description = "Convert books and documents into structured, on-demand agent skills."
+readme = "README.md"
+requires-python = ">=3.9"
+license = { text = "MIT" }
+
+# Tooling config so `ruff check .` and `pytest` locally match CI without flags.
+[tool.ruff]
+target-version = "py39"
+
+[tool.ruff.lint]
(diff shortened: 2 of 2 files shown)


virgiliojr94 (approved):
Approving — clean, coherent with the 3.9 floor (now supported via #34 and guarded in the CI matrix), and the pyproject is metadata + local-tooling only (no build backend, tool stays CLI-first). 

Three minor, non-blocking notes:
- The CI lint passes ruff flags explicitly (`--select E9,F --target-version py310 scripts/ tests/`), so it does not read this pyproject — `[tool.ruff]` here only affects local `ruff check .`, which is exactly the intent. Just note the pyproject says `target-version=py39` while CI still says `py310`; now that 3.9 is the floor, aligning the CI flag to py39 would be the consistent follow-up (E9/F aren't version-sensitive, so not urgent).
- `version = "1.1.0"` will drift on each release — either remember to bump it or drop it since there's no build backend.
- Touches CHANGELOG [Unreleased]; trivial rebase if #36 lands first.

Fork PR, so CI is pending approval — will confirm green before merge. Thanks!

virgiliojr94:
The CHANGELOG conflict is just a section collision: #34 (already merged) added an `## [Unreleased]` section, and this PR adds another one. Resolution is to **merge into the existing `[Unreleased]`** rather than create a second header — put the pyproject bullet under an `### Added` subsection, keeping #34's `### Fixed` / `### Changed` intact. After rebasing on `master`, the top of CHANGELOG.md should read:

```markdown
## [Unreleased]

### Added
- **`pyproject.toml`** declaring `requires-python = ">=3.9"` and centralizing the
  ruff/pytest config so `ruff check .` and `pytest` locally match CI without
  passing flags by hand. Metadata only — no build backend or console scripts; the
  tool stays CLI-first via `python3 scripts/extract.py`.

### Fixed
- **Package now imports on interpreters that evaluate annotations eagerly.** … (from #34)

### Changed
- CI test matrix now includes Python 3.9 … (from #34)
```

`git rebase origin/master`, fix the one CHANGELOG hunk that way, force-push, and it's good. Everything else in the PR is approved. Thanks!

PR #36, opened 2026-06-13
feat: Chinese (CJK) chapter detection + UTF-8 output fix on Windows

wuji-labs:
Two small, self-contained improvements found while converting Chinese-language books with this tool.

## 1. Chinese chapter-heading detection

`detect_structure()` recognized only Latin (`Chapter N`, `Capítulo N`) and Roman-numeral headings, so Chinese sources were reported as **0 chapters**. For example a 15-lecture book whose sections are `第一讲 … 第十五讲`, or Markdown docs with `## 一 · …` headings, got no structure at all.

This adds detection for:
- explicit `第N章 / 第N回 / 第N节 / 第N卷 / 第N篇 / 第N讲` (Arabic or Chinese numerals)
- Markdown headings led by a CJK ordinal: `## 一 · 缘起`, `## 第一讲`

A small Chinese-numeral parser (`一二三` / `十百千` / `〇零` → int) lets these dedupe by number exactly like the existing Arabic/Roman logic (a ToC entry and its body heading still count once).

It is **scoped to CJK numerals**, so Latin/Roman detection is completely unchanged — `## 5 Setup` is still not treated as a chapter (test included).

## 2. UTF-8 output on Windows

`extract.py --check` and `tools/validate_skill.py` raise `UnicodeEncodeError` on Windows consoles that default to a legacy code page (cp936/GBK), because they print the braille attribution banner and the `✓`/`✗` glyphs. Forcing `sys.stdout`/`sys.stderr` to UTF-8 at startup fixes it; no logic changes.

## Tests

6 new CJK chapter-detection tests + a Chinese-numeral-parser test. Full suite: **67 passed**, no regressions.

## Notes

Chapter count is only a hint for the downstream analysis step, so this widens coverage without altering existing behavior on non-CJK input.

PR #44, opened 2026-06-14
feat(extractor): Markdown/AsciiDoc heading detection + multilingual ToC

dex0shubham:
## What & why

`detect_structure()` (`scripts/extractor/utils.py`) feeds the Step 2.5 cost
estimate and Step 3 chapter-mapping, and a zero/false result triggers the
"chapter mapping may miss or duplicate sections" warning. Two gaps were
confirmed by running the current `master` code:

**1. Markdown / AsciiDoc headings were invisible.** The tool supports `.md`,
`.markdown`, `.rst`, `.adoc` (they're in `TEXT_EXTENSIONS`), but chapter
detection only recognized the literal word "Chapter"/"Capítulo", Roman numerals,
or CJK headings. A Markdown book using `# Introduction` / `## Getting Started`
reported **0 chapters**. Measured before this PR:

| Input | `chapters_detected` (before) |
|---|---|
| `# Book Title` + three `## Chapter` headings | 0 |
| three top-level `#` chapter headings | 0 |
| AsciiDoc `= Title` + two `== Section` | 0 |
| `## Chapter 1:` / `## Chapter 2:` | 0 |

**2. Table-of-contents detection was EN/ES/PT only.** Chapter detection already
handles CJK (#36) and Roman numerals (#28), but `has_toc` missed Chinese `目录`,
Japanese `目次`, French `Table des matières`, German `Inhaltsverzeichnis`,
Italian `Indice`, Dutch `Inhoudsopgave`.

### Approach
- **New `_structural_chapter_count()`** counts ATX-style headings (`#`–`######`
  Markdown, `=`–`======` AsciiDoc). It runs **only as a fallback when the numeric
  "Chapter N" scan returns 0**, so prose/PDF books are completely unaffected
  (numeric detection always wins). It skips fenced code blocks (so a `# comment`
  in a sample isn't a chapter) and bare-digit-led titles like `## 5 Setup`
  (preserving the existing `test_cjk_detection_does_not_affect_latin` behavior).
  It picks the shallowest heading depth with ≥2 distinct titles — the real
  chapter level in the common `# Book Title` / `## Chapter` layout.
- **`_TOC_PATTERN`** replaces the inline regex with a readable multilingual
  header list, whole-line-anchored so an inline "the contents of this chapter"
  never matches.

**Scope note:** reStructuredText *underline* headings (`Title` / `=====`) are
intentionally out of scope — they collide with horizontal rules and table
borders. ATX `#`/`=` only. Happy to split this into two PRs (ToC / headings) if
you'd prefer smaller units — the commits are already separated that way.

## Evidence

23 new tests (all green), covering both features and the key regression guards:

- **Structural:** `#`/`##` layouts, all-`#` chapters, AsciiDoc `==` and deeper
  `===` levels, markdown-prefixed `## Chapter N`, fenced-code-block exclusion,
  plain-prose-still-0, and **numeric-wins precedence**
  (`Chapter 1 … ## a/b/c … Chapter 2` → 2, not 5).
- **ToC:** CN (simplified + traditional), JP, FR, DE, IT, NL positives, plus an
  inline-word false-positive guard.

```
$ pytest -q
85 passed
```

Before/after on a Markdown source:

```
# detect_structure("# Book Title\n## Introduction\n## Getting Started\n## Advanced\n")
before: chapters_detected = 0
after:  chapters_detected = 3
```

Precision note: the shorter German/French aliases `inhalt` / `sommaire` were
deliberately **not** added (ordinary section words → false positives); the
unambiguous `inhaltsverzeichnis` / `table des matières` back the DE/FR claims.

## Checklist
- [x] One focused change (structure-detection robustness; commits split by feature)
- [x] Tests added/updated for behavior changes (23 new)
- [x] `ruff check --select E9,F --target-version py310 scripts/ tests/` clean (the CI lint command — passes; pre-existing E402 in the test import block is unrelated and outside CI's selection)
- [x] `pytest -q` green (85 passed)
- [ ] `python3 tools/validate_skill.py SKILL.md` — N/A (SKILL.md unchanged)
- [x] No raw book text shipped; no SKILL.md changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Structure detection now recognizes Markdown/AsciiDoc ATX headings (`#`, `==`)
+  as chapters when no numeric "Chapter N" headings are present, fixing a
+  zero-chapter result for `.md`/`.adoc` sources. Headings inside fenced code
+  blocks are ignored.
+- Table-of-contents detection extended to Chinese, Japanese, French, German,
+  Italian, and Dutch.
+
 ### Security
 - **CI security scanning** — added CodeQL (Python, security-and-quality + weekly
   schedule), Bandit (gates on HIGH severity; reports MEDIUM+ informationally), and
--- a/scripts/extractor/utils.py
+++ b/scripts/extractor/utils.py
@@ -85,6 +85,21 @@ _CN_NUM_CLASS = "〇零一二两三四五六七八九十百千"
 _CN_CHAPTER = re.compile(rf"^\s*第\s*([0-9{_CN_NUM_CLASS}]+)\s*[章回卷节篇讲]")
 _MD_CN_HEADING = re.compile(rf"^#{{1,6}}\s+第?\s*([{_CN_NUM_CLASS}]+)\s*[·、.:：章回卷节篇讲]")
 
+# Table-of-contents header lines across common languages. Anchored to a whole
+# line (^\s*X\s*$) so an inline "the contents of this chapter" never matches.
+_TOC_HEADERS = (
+    "table of contents", "contents", "índice", "sumário",   # EN / ES / PT
+    "目录", "目錄", "目次",                                   # Chinese / Japanese
+    "table des matières",                                   # French
+    "inhaltsverzeichnis",                                   # German
+    "indice", "sommario",                                   # Italian (no accent — distinct from índice above)
+    "inhoudsopgave",                                        # Dutch
+)
+_TOC_PATTERN = re.compile(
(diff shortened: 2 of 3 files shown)


dex0shubham:
Thanks for maintaining this — the discovery-loop-tax framing is a genuinely clever approach. 👋

Quick heads-up: this is my first PR to the repo, so the Actions workflows are waiting on a maintainer's **"Approve and run"** before CI will report. For what it's worth, I've reproduced both locally: `pytest -q` → 85 passed, and the exact CI lint (`ruff check --select E9,F --target-version py310 scripts/ tests/`) is clean.

The change extends `detect_structure()` along the same path as the Roman-numeral (#28) and CJK (#36) work — Markdown/AsciiDoc headings as a *fallback-only* signal (numeric "Chapter N" always wins, so existing books are untouched) plus multilingual ToC headers. The two features are in separate commits, so I'm happy to split them into two PRs if you'd rather review them independently.

No rush at all — just flagging the CI gate so it's not silently stuck. Thanks for taking a look!

PR #50, opened 2026-06-16
fix: log unexpected exceptions in parsers to stderr

addy790:
fix(parsers): log unexpected exceptions to stderr instead of silently returning None

## What & why
Every parser in `scripts/extractor/parsers/` had a broad `except Exception: return None` that silently swallowed real errors — corrupt files, encoding issues, genuine bugs — making them impossible to debug. This adds a one-line warning to stderr for each unexpected failure while keeping the fallback chain exactly as-is.

## Evidence
No behavior change — extractors still return `None` on failure so the chain continues. Before this, a corrupt PDF gave you a silent `None`. After this, you get:

 [warn] extract_with_pypdf2 failed: PdfReadError: EOF marker not found

and the run still proceeds to the next fallback.

## Checklist
- [x ] One focused change
- [x ] `ruff check .` clean
- [x ] `pytest -q` green
- [ ] `python3 tools/validate_skill.py SKILL.md` passes (if SKILL.md changed)
- [x ] `CHANGELOG.md` updated under `## [Unreleased]`
- [x ] No raw book text shipped; no net SKILL.md bloat without justification


virgiliojr94 (commented):
Thanks for tackling this, the behavior is exactly right: unexpected exceptions now surface on stderr while the fallback chain still returns `None` and continues. Verified locally, 87 tests green. Two things before merge:

1. **Add the test.** The issue asks for a test that proves a deliberately broken input still falls through to `None` without breaking the chain. Right now there is no test covering the new behavior, so a future refactor could silently drop the warning or change the control flow and CI would not catch it. A small one is enough, for example: monkeypatch one extractor to raise, capture `capsys`, and assert (a) the function returns `None` and (b) a `[warn] ... failed:` line was written to stderr.

2. **Fix the indentation.** In `calibre.py`, `pdf.py` (pypdf2 / pdfminer / docling), and `rtf.py` the `print(...)` is indented 8 spaces past the `except`, e.g.:
   ```python
   except Exception as e:
           print(...)   # over-indented
   ```
   It runs, but it is inconsistent with the rest of the file (the docx/epub/text edits are correctly at 4 spaces). Please normalize to 4 so it reads clean.

One tiny nit: the CHANGELOG bullet reads as a run-on, a period or dash after "`None`" would fix it.

Everything else is good. Ping me when updated and I will merge.

addy790:
Added the regression test and fixed the indentation issue in extract_with_pdftotext, all 88 tests passing locally. Ready for review.

PR #74, opened 2026-07-22
feat(security): scan generated skills

rhein1:
## Summary (Required)

Adds a deterministic, standard-library advisory scanner for generated skills. The conversion workflow now scans generated SKILL.md and direct chapters/*.md files before final reporting or publication, returning non-zero when known prompt-authority, invisible-Unicode, frontmatter-widening, or exfiltration-shaped patterns are present.

## Type of change (Required)

- [ ] Bug fix
- [x] Feature
- [ ] Performance
- [ ] Refactor
- [ ] Docs only
- [x] Security
- [ ] Chore / CI / tooling
- [ ] Breaking change

## What it does (Required)

- **Adds:** tools/scan_generated_skill.py, a dependency-free advisory scanner with bounded file count and byte limits, symlink rejection, strict UTF-8 handling, escaped diagnostics, and non-zero findings.
- **Adds:** eight focused tests covering clean output, scanner scope, prompt-control tokens, invisible Unicode, authority-widening frontmatter, exfiltration-shaped content, redacted CLI output, control-character escaping, and oversized files.
- **Improves:** SKILL.md now runs the scanner after validation in both creation and update flows, before final reporting or publication.

## Motivation & context (Required)

Refs #73. This addresses only Gap 2, generated-skill output validation. Gap 1, extraction-time invisible-Unicode scrubbing, remains intentionally out of scope so this PR does not overclaim closure of the issue.

## How it works (Required)

The scanner parses generated SKILL.md frontmatter, then scans that file and direct chapters/*.md files with explicit deterministic rules. It does not rewrite generated content or echo matched attacker-controlled text. Findings are advisory but fail the command with a non-zero exit code so a human can review before installation or sharing.

The scope is intentionally narrow: only the generated skill entrypoint and direct chapter files are scanned. Build metadata, tests, and arbitrary nested files are excluded to avoid treating repository internals as generated skill content.

## Evidence (Required)

- **Tests:** The complete suite passes on every CI-supported interpreter: Python 3.9, 3.10, 3.11, 3.12, and 3.13, with 151 passing tests on each.
- **Before / after:** Before this change, structural validation did not inspect generated content for prompt-authority or exfiltration patterns. After this change, planted payloads and frontmatter widening are covered by tests and cause the scanner to return non-zero; the repository's clean skill returns zero.

    Python 3.9: 151 passed
    Python 3.10: 151 passed
    Python 3.11: 151 passed
    Python 3.12: 151 passed
    Python 3.13: 151 passed
[...]

Issue #73, opened 2026-07-21
Harden the doc→agent→skill pipeline against indirect prompt injection (invisible-Unicode scrub + generated-skill output scan)


virgiliojr94:
## Summary

book-to-skill is a **document → LLM context supply chain**: an untrusted document flows through our parsers into an agent's prompt, and the agent then writes a *new* skill that gets loaded into other agents later. That makes two indirect-prompt-injection surfaces worth hardening that **no open PR currently covers**. Opening this to track them and invite contributors.

This is defense-in-depth, not a live exploit report — our subprocess call sites are already list-arg + `os.path.abspath` (thanks to #54), DOCX XXE/Billion-Laughs is guarded, and there's no doc-controlled network fetch. The gaps below are about the *content* that reaches the agent and the skill we emit.

## Threat model

```
[1] untrusted doc (downloaded from anywhere)
      │  ← attacker-controlled
      ▼
[2] parsers → full_text.txt + metadata.json          (α: code hardening — mostly done)
      ▼
[3] agent reads full_text.txt INTO its prompt         (β: instruction/data boundary)
      ▼
[4] agent writes SKILL.md + chapters/ + glossary…     (γ: the generated skill can be poisoned)
      ▼
[5] generated skill loaded into future agents, repeatedly   (persistence / propagation)
```

The `[2]→[3]` arrow is the trust boundary: everything to its left is *data*, and our SKILL.md is the only trusted *instruction* source. See [OWASP LLM01:2025 — Prompt Injection](https://genai.owasp.org/llmrisk/llm01-prompt-injection/) (indirect/doc-borne injection is co-equal to direct).

## Gap 1 — Extraction does not strip invisible/zero-width Unicode (surface α→β)

Extraction is faithful by design, so hidden payloads survive verbatim into the agent's context:

- Zero-width chars: `U+200B`, `U+200C`, `U+200D`, `U+FEFF`
- **Unicode tag block `U+E0000–U+E007F`** — renders invisible, models still read it ([Cisco](https://blogs.cisco.com/ai/understanding-and-mitigating-unicode-tag-prompt-injection), [Trend Micro](https://www.trendmicro.com/en_us/research/25/a/invisible-prompt-injection-secure-ai.html))
- HTML/EPUB text hidden via CSS (`display:none`, `visibility:hidden`, off-screen positioning) — our stdlib HTML parser drops `script`/`style`/comments but does not evaluate CSS, so hidden text is still emitted

[...]

PR #75, opened 2026-07-22
fix(security): scrub invisible Unicode from extraction

rhein1:
## Summary (Required)

Removes zero-width U+200B/U+200C/U+200D/U+FEFF characters and the Unicode tag block U+E0000-U+E007F from every parser result before metrics or full_text.txt are produced. Extraction reports the number removed and rejects a source that contains no visible content after sanitization.

## Type of change (Required)

- [x] Bug fix
- [ ] Feature
- [ ] Performance
- [ ] Refactor
- [ ] Docs only
- [x] Security
- [ ] Chore / CI / tooling
- [ ] Breaking change

## What it does (Required)

- **Fixes:** Invisible Unicode from untrusted documents previously survived extraction and entered the agent context unchanged.
- **Adds:** A dependency-free sanitize_extracted_text boundary shared by every supported parser path.
- **Improves:** Character, word, token, and structure metrics now describe the sanitized text actually written to full_text.txt.

Refs #73. This addresses only Gap 1, extraction-time invisible-Unicode scrubbing. Gap 2, generated-skill output scanning, is independently proposed in #74, so this PR does not claim to close the whole issue by itself.

## How it works (Required)

All format-specific parsers already converge in extract_single_file. The new sanitizer runs once at that boundary, before metrics and structure detection. It removes only the exact zero-width code points and Unicode tag range named in #73, counts removals, and leaves ordinary CJK, combining marks, emoji, whitespace, and visible text unchanged.

A source that becomes blank after sanitization raises ExtractionError rather than emitting an empty context artifact.

## Evidence (Required)

- **Tests:** Five focused tests cover all four zero-width code points, tag-range boundaries and a middle value, byte-for-byte preservation of normal multilingual text, sanitized direct extraction and metrics, rejection of invisible-only content, and the final full_text.txt write path.
- **Before / after:** Before this change, planted U+200B and U+E0069 code points remained in extracted text. After this change, the visible sentence remains, both hidden code points are absent, metrics use the sanitized length, and stderr reports a removal count.

    Python 3.9: 148 passed
    Python 3.10: 148 passed
    Python 3.11: 148 passed
    Python 3.12: 148 passed
    Python 3.13: 148 passed
    Ruff: All checks passed
    Bandit high severity / medium confidence gate: passed
    SKILL.md validator: passed with the existing soft size warning

## Screenshots / output

    [security] removed 2 invisible Unicode code point(s) from source.txt

The end-to-end regression confirms full_text.txt contains Policy text. and none of the planted invisible code points.

## Checklist (Required)

- [x] Branch is rebased on the latest master and has no merge conflicts
- [x] pytest -q is green on a clean checkout
- [x] ruff check . is clean
- [x] python3 tools/validate_skill.py SKILL.md passes (SKILL.md is unchanged; one existing soft size warning remains)
- [x] CHANGELOG.md updated under ## [Unreleased]
- [x] No raw book text shipped; SKILL.md is unchanged

## Breaking changes & back-compat

Visible extracted content and existing command-line arguments are unchanged. Documents that contain the targeted invisible code points now lose only those code points; invisible-only documents fail extraction instead of producing a hidden-content artifact.

## Notes for reviewers

CSS-hidden HTML text is not addressed here. This PR deliberately implements the deterministic Unicode acceptance criteria from #73 without attempting semantic or CSS-layout sanitization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   screen instead of being buried mid-page.
 
 ### Security
+- **Invisible-Unicode extraction hardening** — every parser result now removes
+  zero-width U+200B/U+200C/U+200D/U+FEFF characters and the Unicode tag block
+  U+E0000-U+E007F before metrics or `full_text.txt` are produced, reports the
+  removal count, and rejects sources containing no visible content after the scrub.
 - **DOCX XXE / Billion Laughs hardening** — the DOCX extractor now scans the
   archive and rejects any XML part that declares a DTD or entities before
   parsing, blocking XML external-entity and entity-expansion attacks (#53, #54).
--- a/book_to_skill/sanitize.py
+++ b/book_to_skill/sanitize.py
@@ -1,0 +1,24 @@
+from __future__ import annotations
+
+
+_ZERO_WIDTH_CODEPOINTS = frozenset({0x200B, 0x200C, 0x200D, 0xFEFF})
+_TAG_BLOCK_START = 0xE0000
+_TAG_BLOCK_END = 0xE007F
+
+
+def sanitize_extracted_text(text: str) -> tuple[str, int]:
+    """Remove invisible code points used for document-borne prompt injection."""
+    kept: list[str] = []
+    removed = 0
+
+    for character in text:
+        codepoint = ord(character)
+        if (
+            codepoint in _ZERO_WIDTH_CODEPOINTS
+            or _TAG_BLOCK_START <= codepoint <= _TAG_BLOCK_END
+        ):
+            removed += 1
+            continue
+        kept.append(character)
+
+    return "".join(kept), removed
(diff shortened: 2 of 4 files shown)


virgiliojr94:
@rhein1 this is exactly what #73 Gap 1 asked for — thank you. Stripping only the named zero-width points + the U+E0000–E007F tag block, applied once at the `extract_single_file` convergence point *before* metrics and structure detection, is the right placement, and rejecting an invisible-only source instead of emitting an empty artifact is the detail I was hoping for. The byte-for-byte preservation test on `Café` + combining marks + CJK + emoji is what makes it safe to merge.

One FYI, not a blocker: stripping U+200D (ZWJ) will also split ZWJ emoji ligatures (👨‍👩‍👧 → 👨👩👧). That's an acceptable trade — U+200D is a real injection carrier and book text rarely relies on ZWJ sequences — but worth a one-line comment in `sanitize.py` noting the known cosmetic effect so a future reader doesn't treat it as a bug.

Merging. Clean, tested, scoped honestly to Gap 1. Nice work.

PR #76, opened 2026-07-25
Decode Cyrillic / non-Latin fonts that pdftotext silently drops

leval907:
## Problem

On many digitally-generated PDFs whose fonts are subsetted with no ToUnicode CMap, `pdftotext` (and `pypdf`) drop non-Latin glyphs to punctuation. A Russian book extracts as bare `", . , ,"`. Since `pdftotext` runs first and its non-empty **but garbage** output wins, the fallback chain never runs and the entire book comes out unreadable. Cyrillic, Greek, and similar scripts are affected; Latin text is unaffected.

## Fix

- **`parsers/pdf.py`**: add `extract_with_pdfplumber()` — pdfplumber decodes these fonts correctly — plus `alpha_ratio()` / `looks_corrupt()`, a cheap gate that flags punctuation-dominated output (`alpha_ratio < 0.35`, calibrated on real garbage ~0.02–0.11 vs clean text ~0.51–0.96).
- **`utils.py`**: the PDF text path is now a gated loop `pdftotext → pdfplumber → pypdf → pdfminer` — the first non-empty **and non-corrupt** result wins. Docling (technical mode) is gated the same way. If every extractor looks corrupt (an unusual scan), the least-bad result is kept instead of hard-failing.
- **`config.py` / `dependencies.py`**: register `pdfplumber`; `--check` and the install prompt surface it (flagged best for Cyrillic / non-Latin fonts).

## Compatibility

Backward compatible. For Latin PDFs `pdftotext` still wins on the first try; `pdfplumber` is optional (already a common dependency) and only used when `pdftotext` output is empty or garbled.

## Testing

Verified on a 542-page Russian PDF: `pdftotext` alpha_ratio **0.11** (flagged corrupt) → falls through to `pdfplumber` alpha_ratio **0.92** (clean Cyrillic). Gate unit-checked: flags real garbage, passes clean prose, ignores empty/short strings.

### DIFF
--- a/book_to_skill/config.py
+++ b/book_to_skill/config.py
@@ -25,6 +25,7 @@ SUPPORTED_EXTENSIONS = {
 
 PYTHON_DEPENDENCIES = {
     "docling": "docling",
+    "pdfplumber": "pdfplumber",
     "pypdf": "pypdf",
     "pdfminer": "pdfminer.six",
     "ebooklib": "ebooklib",
--- a/book_to_skill/dependencies.py
+++ b/book_to_skill/dependencies.py
@@ -14,11 +14,11 @@ from book_to_skill.config import PYTHON_DEPENDENCIES, HTML_EXTENSIONS
 DEPENDENCY_GROUPS = [
     {
         "label": "PDF (text-heavy)",
-        "modules": ["pypdf", "pdfminer"],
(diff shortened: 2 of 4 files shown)


leval907:
Closing — keeping this as a local adaptation, not upstreaming.

Issue #82, opened 2026-07-28
Add Korean chapter-heading detection (제N장) to _chapter_number()


imchul-oss:
## Summary

`_chapter_number()` covers six Latin-script languages, Roman numerals, Chinese (`第N章`) and Thai (`บทที่`), but has no Korean pattern. Korean books and documents number chapters as `제N장` — Hangul `제`, not the Han `第` — so `_CN_CHAPTER` never matches and Korean sources always fall through to the structural fallback.

Split out of #81 as requested there: that one is a defect in the fallback, this one is the missing pattern that sends Korean to the fallback in the first place.

## Evidence

Measured against a local mirror of the Korean statute corpus (3,030 statutes / 5,728 Markdown files), random sample of 600 files: **34% contain `제N<classifier>` headings**. Classifier frequency across the sample:

| classifier | meaning | occurrences |
|---|---|---:|
| `장` | chapter | 1,323 |
| `절` | section | 267 |
| `관` | subsection | 11 |
| `편` | part | 5 |

`장` dominates, as `章` does in Chinese.

## Proposed pattern

Mirroring the shape of the existing `_CN_CHAPTER` / `_MD_CN_HEADING` pair:

```python
# Korean chapter headings: "제1장 총칙", "## 제4장 근로시간과 휴식", "제6장의2 …".
# 제 + Arabic numeral + a classifier (장 chapter / 편 part / 절 section / 관 subsection),
# with an optional "의N" branch suffix that Korean statutes use for inserted chapters
# (제6장의2). Modern Korean numbers chapters with Arabic digits, so unlike the Chinese
# branch no numeral composition is needed. Optional Markdown "#" prefix so "## 제1장"
# is recognized in converted ebooks.
_KO_CHAPTER = re.compile(
    r"^\s*(?:#{1,6}\s+)?제\s*(\d+)\s*[장편절관](?:\s*의\s*\d+)?(?:\s*$|[.:\-]|\s+\S)"
)
```

The trailing group is the Korean analogue of `_HEADING_TAIL`: Korean has no letter case, so the existing "capitalized title word" test does not transfer. Requiring end-of-line, punctuation, or whitespace-then-content is what separates a heading from a prose cross-reference, because Korean particles attach directly to the noun (`제5장에서`, `제2장의`) with no intervening space.

## Measured precision / recall

Same corpus, random sample of 400 files, ground truth = lines that are Markdown headings beginning with `제N<classifier>`:

```
[...]

PR #83, opened 2026-07-28
fix(extractor): exclude source banners from chapter detection

### LINKED ISSUE #81: Extractor's own SOURCE banner is parsed as a setext heading, collapsing chapters_detected to 2 on structural-fallback books
## Summary

The `SOURCE: ...` banner that `extract.py` prepends to `full_text.txt` is itself parsed as a setext heading, which corrupts `_structural_chapter_count()` and makes the consolidated `chapters_detected` collapse to `2` for any book that relies on the structural fallback.

`metadata.json` ends up carrying both the right and the wrong number: the per-source record has the correct count, while the top-level value — the one the agent reads in Step 3 — has `2`.

This is language-independent and reproduces on plain English Markdown.

## Reproduction

`sample.md` — five `##` sections, no `Chapter N` headings:

```markdown
# The Pragmatic Widget

Intro paragraph.

## Foundations

Body text for foundations.

## Design Rules

Body text for design rules.

## Trade-offs

Body text for trade-offs.

## Operating Model

Body text for operating model.

## Closing

Body text for closing.
```

```console
$ python scripts/extract.py sample.md --mode text
   Sources : 1 processed
   Words   : 46
   Chapters: 2 detected overall
```

Expected `5`, got `2`.

The resulting `metadata.json` contradicts itself:

```json
"sources": [ { "chapters_detected": 5, ... } ],
"chapters_detected": 2,
"chapter_headings_sample": []
```

## Root cause

In `book_to_skill/utils.py`, `_structural_chapter_count()` treats a line of `=` as a setext underline and registers the line above it as a depth-1 heading:

```python
if (_SETEXT_UNDERLINE.match(s) and prev
        and not _SETEXT_UNDERLINE.match(prev) and len(s) >= len(prev)):
    depth = 1 if s[0] == "=" else 2
    levels.setdefault(depth, set()).add(prev.lower())
```

The consolidated text begins with the extractor's own banner:

```
================================================================================
SOURCE: sample.md (Path: /tmp/sample.md)
================================================================================
```

The second rule line has `prev = "SOURCE: sample.md (Path: /tmp/sample.md)"`, so that line is added as a phantom depth-1 title. Together with the book's real `# The Pragmatic Widget` this puts **two** distinct titles at depth 1, and the function returns at the shallowest depth with `>= 2` titles — never descending to depth 2 where the five real chapters live.

Isolated at function level:

```python
from book_to_skill.utils import _structural_chapter_count as c
c(md)                                    # 5   (raw file)
c(banner + md)                           # 2   (as consolidated)
c(strip_rule_lines(banner + md))         # 5   (banner rule lines removed)
```

### Why it looks intermittent

The guard is `len(s) >= len(prev)`, and the banner rule is a fixed 80 characters. So the bug fires only when the `SOURCE:` line is **80 characters or shorter** — i.e. it depends on the length of the input path:

| `SOURCE:` line length | `chapters_detected` |
|---|---|
| 40 | 2 (wrong) |
| 70 | 2 (wrong) |
| 116 | 5 (correct) |

A deeply nested input path masks the bug; a short one triggers it.

## Impact

Any book without explicit `Chapter N` headings takes the structural fallback, so this hits the exact population the fallback exists to serve. That includes the two books the README itself names as non-segmenting — Pro Git (section titles) and Moby-Dick (chapter titles / roman numerals) — and every Markdown/AsciiDoc/RST source folded in via multi-file input.

Downstream, Step 3 hands the agent `chapters_detected: 2` plus an empty `chapter_headings_sample`, so chapter planning starts from a wrong structure, and the Step 7 loop generates far fewer chapter files than the source has sections.

## Suggested direction

Not prescribing an implementation, but the cleanest fix is probably to run structure detection on the per-source text before the banner is prepended (the per-source value is already correct), or to have `detect_structure()` skip the banner block, rather than to tune the setext heuristic.

Happy to send a PR if you'd like a particular approach.

---

Found while evaluating the tool on Korean-language sources. Separately, `_chapter_number()` has no Korean pattern — Korean statutes and books use `제N장`, and the Chinese branch requires the Han `第` so Hangul `제` never matches. That is an enhancement rather than a defect, so I have kept it out of this issue; happy to open it separately if it is welcome.

hydraxman:
## Summary (Required)

Exclude generated `SOURCE:` boundary banners from consolidated structure detection so short source paths no longer collapse structural chapter counts to 2. The emitted `full_text.txt` remains unchanged.

## Type of change (Required)

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)

- **Fixes:** consolidated chapter detection no longer parses the extractor's generated `SOURCE:` banners as setext headings.
- **Improves:** top-level `chapters_detected` now agrees with the per-source count for structural-fallback books regardless of source-path length.

The generated banner uses 80-character `=` rules. For a short source path, the second rule turns the preceding `SOURCE:` line into a phantom depth-1 setext heading. Combined with a real H1 title, that makes `_structural_chapter_count()` return 2 before reaching the actual chapter level.

Fixes #81

The extraction flow still writes the same banner-delimited `full_text.txt` and still includes those bytes in output metrics. Only structure detection receives a separate view built by joining the already-extracted source texts without generated presentation banners.

This preserves cross-source numeric chapter deduplication and table-of-contents detection without coupling the general parser to the current banner format. The alternative—special-casing banners inside `detect_structure()`—would make parser behavior depend on generated output formatting.

## Evidence (Required)

- **Tests:** added an integration regression that runs the real extraction flow for both short and long reported source paths, verifies per-source and top-level counts remain 5, and verifies the source banner is still emitted.
- **Before / after:** the short-path case fails on `master` with a top-level count of 2 and passes after the fix with 5.

```text
Before (master c6bc1b7):
sources[0].chapters_detected = 5
chapters_detected = 2
assert 2 == 5

After:
$ python -m pytest -q tests/test_book_to_skill.py::TestBatchResilience::test_source_banner_does_not_change_structural_chapter_count
2 passed in 0.09s

$ python -m pytest -q
165 passed in 0.15s

$ ruff check .
All checks passed!
```

```text
Before: Chapters: 2 detected overall
After:  Chapters: 5 detected overall
```

- [x] `ruff check .` is clean

Korean heading support from #82 is intentionally out of scope; this fixes the language-independent fallback path only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `PyPDF2` is end-of-life and no longer receives security fixes (#54).
 
 ### Fixed
+- Consolidated chapter detection now analyzes extracted source text without the generated
+  `SOURCE:` boundary banners, preventing those banners from becoming phantom setext headings
+  and collapsing `chapters_detected` to 2 for short source paths (#81).
 - **`Chapter I.` — a chapter word followed by a Roman numeral — is now detected.** It
   matched neither existing pattern (`_EXPLICIT_CHAPTER` required Arabic digits after the
   chapter word; `_ROMAN_HEAD` required the numeral to start the line), so books using
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -646,8 +646,11 @@ def main():
     total_words = len(consolidated_text.split())
     total_tokens = estimate_tokens(consolidated_text)
     
-    # Detect structure on consolidated text
-    consolidated_structure = detect_structure(consolidated_text)
+    # Detect structure from source content only. The generated SOURCE banners in
+    # full_text.txt use rows of "=", which can otherwise become phantom setext
+    # headings and make the result depend on the source-path length.
+    structure_text = "\n\n".join(src["text"] for src in extracted_sources)
+    consolidated_structure = detect_structure(structure_text)
     
     metadata = {
         "source_file": "Consolidated from multiple sources" if len(extracted_sources) > 1 else extracted_sources[0]["source_file"],
--- a/tests/test_book_to_skill.py
+++ b/tests/test_book_to_skill.py
@@ -366,6 +366,49 @@ class TestBatchResilience:
         meta = json.loads(out_meta.read_text(encoding="utf-8"))
         assert meta["total_sources"] == 1
 
+    @pytest.mark.parametrize(
+        "reported_source",
+        ["/x/sample.md", "/deep/" + ("nested/" * 12) + "sample.md"],
+    )
+    def test_source_banner_does_not_change_structural_chapter_count(
+        self, tmp_path, monkeypatch, reported_source
+    ):
+        """The generated SOURCE banner must not become a phantom setext heading."""
+        source = _make_md_file(
+            tmp_path / "sample.md",
+            "# The Pragmatic Widget\n\n"
+            "## Foundations\n\nBody.\n\n"
+            "## Design Rules\n\nBody.\n\n"
+            "## Trade-offs\n\nBody.\n\n"
+            "## Operating Model\n\nBody.\n\n"
+            "## Closing\n\nBody.\n",
+        )
+        out_dir = tmp_path / "output"
+        out_text = out_dir / "full_text.txt"
+        out_meta = out_dir / "metadata.json"
+        real_extract = extract_single_file
+
+        def extract_with_reported_source(*args, **kwargs):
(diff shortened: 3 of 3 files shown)


virgiliojr94:
This is the right fix. Running `detect_structure` on the source text *before* the generated `SOURCE:` banners are prepended is exactly it — the `=`-row banner was turning into a phantom setext heading and making `chapters_detected` depend on the source path length. The parametrized test (short vs deeply-nested path) proving stability is the detail I wanted. Merging — closes #81. Thanks @hydraxman.

PR #69, opened 2026-07-01
feat: detect Korean chapter headings and standalone numeric dividers

simartkrom:
## Problem

Korean books and lecture notes are reported as **0 chapters** by `detect_structure()`, because chapter detection only recognizes Latin (`Chapter N`), Roman (`I: Title`), and CJK (`第N章`, `## 一`) heading styles. Real-world Korean sources use `제N장` markers and standalone `01.` dividers, so their chapter mapping in Step 3 falls back to a heading scan that misses everything.

Tested on a real 102-page Korean ebook: **0 → 8 chapters detected** after this change.

## What this adds

Three Korean-common heading styles in `_chapter_number()`:

1. **`제N장 / 제N화 / 제N부 / 제N편 / 제N강 / 제N과`** — the `제` prefix (Hangul *je-*) is an unambiguous chapter marker, mirroring the existing `第N章` handling.
2. **bare `N강 / N화 / N부`** lecture / episode / part units, gated by a Korean-aware heading tail (`_KO_HEADING_TAIL`) so a unit glued to a particle (`4강은`, `3화보다` — no space) is treated as prose and rejected. **`N장` is deliberately excluded** — `장` is also the everyday counter for flat sheets (`사진 10장` = "10 photos"), which would false-match prose.
3. **standalone `N.` divider lines** (e.g. `01.` on its own line with the title on the following line — very common in Korean publishing). Requiring the *whole line* to be only `N.` keeps numbered list items (`1. Buy milk`), step labels (`1단계`), and years (`2025.`) out.

Arabic + full-width digits only (Sino-Korean numerals like `제삼장` are rare and skipped). **Existing Latin / Roman / CJK detection is unchanged.**

## Tests

Adds 5 tests to `TestStructureDetection` covering positives (제N장, N강, `01.` dividers) and the excluded false-positives (sheet counters, particle-glued units, list items, step labels, years). Existing English/Chinese detection tests still pass.

## Notes

- No new dependencies; pure regex additions.
- Motivated by real usage converting a Korean ebook into a skill; happy to adjust naming or scope (e.g. gate the bare `N.` divider more tightly) if you prefer.

PR #84, opened 2026-07-29
fix: add Korean chapter-heading detection (제N장) to _chapter_number()

### LINKED ISSUE #82: Add Korean chapter-heading detection (제N장) to _chapter_number()

varunsahni18:
## Summary (Required)

Adds Korean chapter-heading detection to `_chapter_number()`. Korean books and documents number chapters as `제N장` using Hangul `제` (not the Han `第`), so the existing Chinese branch never matches and Korean sources always fall through to the structural fallback, which produces incorrect chapter counts.

## Type of change (Required)

## What it does (Required)

- **Fixes:** Korean chapter headings (`제N장`, `제N편`, `제N절`, `제N관`) were not detected by `_chapter_number()`, causing Korean books to fall back to structural heuristics and produce incorrect chapter counts.
- **Adds:** Korean language support to the numeric chapter parser, alongside existing Chinese and Thai branches.

Closes #82

Korean is a major language for technical books and legal documents. Without this fix, any Korean PDF or ebook processed by book-to-skill would have its chapters misidentified, leading to broken skill generation.

The regex mirrors the shape of the existing `_CN_CHAPTER` / `_TH_CHAPTER` patterns:

```python
_KO_CHAPTER = re.compile(
    r"^\s*(?:#{1,6}\s+)?제\s*([0-9]+)\s*[장편절관](?:\s*의\s*[0-9]+)?(?:\s*$|[.:\-]|\s+\S)"
)
```

Key design decisions:
- **Arabic digits only**: Modern Korean numbers chapters with Arabic digits, so no numeral composition is needed (unlike the Chinese branch)
- **Tail check**: Korean has no letter case, so the existing `_HEADING_TAIL` test does not transfer. Requiring end-of-line, punctuation, or whitespace-then-content separates headings from prose cross-references where Korean particles attach directly (`제5장에서`, `제2장의`)
- **`조` excluded**: Articles are not chapter classifiers and would produce an order of magnitude more false positives

## Evidence (Required)

- **Tests:** 7 new test cases covering: basic detection, Markdown prefix, inserted-chapter suffix, article exclusion (`제N조`), prose cross-reference rejection, ToC dedup, and other classifiers (`편`/`절`/`관`)
- **Before / after:**

```
# Before: Korean chapter headings not detected
>>> _chapter_number("제1장")
None

# After: Korean chapter headings correctly identified
>>> _chapter_number("제1장")
1
>>> _chapter_number("## 제6장의2")
6
```

- **Corpus validation:** Pattern validated against Korean statute corpus (3,030 statutes / 5,728 Markdown files): precision 0.999, recall 1.000

- [x] `ruff check .` is clean

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -117,7 +117,7 @@ _TH_CHAPTER = re.compile(
 # a heading from a prose cross-reference, because Korean particles attach directly
 # to the noun ("제5장에서", "제2장의") with no intervening space.
 _KO_CHAPTER = re.compile(
-    r"^\s*(?:#{1,6}\s+)?제\s*(\d+)\s*[장편절관](?:\s*의\s*\d+)?(?:\s*$|[.:\-]|\s+\S)"
+    r"^\s*(?:#{1,6}\s+)?제\s*([0-9]+)\s*[장편절관](?:\s*의\s*[0-9]+)?(?:\s*$|[.:\-]|\s+\S)"
 )
 
 # Table-of-contents header lines across common languages. Anchored to a whole
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -104,6 +104,22 @@ _TH_CHAPTER = re.compile(
     rf"^\s*(?:#{{1,6}}\s+)?(?:บทที่|ตอนที่|ภาคที่|บท|ตอน|ภาค)\s*([0-9{_TH_DIGITS}]+)\b"
 )
 
+# Korean chapter headings: "제1장 총칙", "## 제4장 근로시간과 휴식", "제6장의2 …".
+# 제 + Arabic numeral + a classifier (장 chapter / 편 part / 절 section / 관
+# subsection), with an optional "의N" branch suffix that Korean statutes use for
+# inserted chapters (제6장의2). Modern Korean numbers chapters with Arabic digits,
+# so unlike the Chinese branch no numeral composition is needed. Optional Markdown
+# "#" prefix so "## 제1장" is recognized in converted ebooks.
+#
+# The trailing group is the Korean analogue of _HEADING_TAIL: Korean has no letter
+# case, so the existing "capitalized title word" test does not transfer.
+# Requiring end-of-line, punctuation, or whitespace-then-content is what separates
+# a heading from a prose cross-reference, because Korean particles attach directly
+# to the noun ("제5장에서", "제2장의") with no intervening space.
+_KO_CHAPTER = re.compile(
+    r"^\s*(?:#{1,6}\s+)?제\s*(\d+)\s*[장편절관](?:\s*의\s*\d+)?(?:\s*$|[.:\-]|\s+\S)"
+)
+
 # Table-of-contents header lines across common languages. Anchored to a whole
 # line (^\s*X\s*$) so an inline "the contents of this chapter" never matches.
 _TOC_HEADERS = (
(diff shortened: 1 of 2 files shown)


virgiliojr94:
Excellent — template-complete, rebased on current `master`, and the corpus validation (precision 0.999 / recall 1.000 on ~3k statutes) is exactly the evidence I want to see. Requiring the `제` prefix makes it safe by construction — `사진 10장` can't collide by design. Approving the CI run now; I'll merge once it's green. Closes #82.

Optional follow-up (not a blocker): #69 also covered ebook/lecture forms — bare `N강`/`N화` units, the `화`/`강`/`과` classifiers, and full-width digits. If you'd like to add `화강과` to the class + full-width digit support in a follow-up, this would cover non-statute Korean too. Happy to merge this as-is and take that separately.

PR #94, opened 2026-08-02
chore: untrack stale compiled bytecode

tjaiyen:
## Summary (Required)

`scripts/__pycache__/extract.cpython-313.pyc` is tracked in git even though `.gitignore` lists both
`*.pyc` and `__pycache__/` — ignore rules don't untrack files that were already committed. This
removes it and adds a test that fails if compiled bytecode is ever tracked again.

## Type of change (Required)

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [x] 🧹 Chore / CI / tooling

## What it does (Required)

- **Fixes:** a stale compiled artifact shipped in the repo. It entered in `d88e69e`
  ("feat: add EPUB support") and has been tracked ever since. Root cause: `.gitignore` was correct,
  but adding an ignore rule does not untrack an already-committed file.

Closes n/a — no issue; found while reviewing the repo before use.

To be clear about severity: **the file was inert, and this is not a vulnerability report.** It uses a
timestamp-based `.pyc` header recording an 11,321-byte source, while `scripts/extract.py` is now 871
bytes, so CPython's validation always failed and it recompiled from source. It never executed.

It's still worth removing for two reasons:

1. It matches **no committed revision** of `scripts/extract.py`. I compiled all 13 historical
   revisions and compared structurally (line-number-insensitive); none reproduce it. It was compiled
   from an uncommitted mid-edit working state and swept in with the surrounding commit. So the one
   file in the repo that isn't reviewable in a diff is also the one file that corresponds to no
   reviewable source.
2. This project generates agent skills that people load into Claude Code / Copilot / Amp. Shipping
   opaque bytecode is a bad look for a tool whose output is meant to be trusted — and the repo already
   takes this seriously elsewhere (`sanitize.py`, `tools/scan_generated_skill.py`).

`git rm --cached` on the artifact, plus `tests/test_repo_hygiene.py` asserting that `git ls-files`
returns no `.pyc` or `__pycache__/` paths.

The test skips rather than fails when it can't see a git checkout (installed sdist, `git` missing,
non-zero exit), so it won't break packaging or offline test runs.

## Evidence (Required)

- **Tests:** `tests/test_repo_hygiene.py::test_no_compiled_bytecode_is_tracked` asserts no tracked
  `.pyc` / `__pycache__/` paths, and names offenders in the failure message.

**The guard was verified to actually fire**, not just to pass. Re-staging the artifact with
`git add -f` makes it fail:

```
$ git add -f scripts/__pycache__/extract.cpython-313.pyc
$ pytest tests/test_repo_hygiene.py -q
E       AssertionError: compiled bytecode is tracked in git: scripts/__pycache__/extract.cpython-313.pyc — remove with `git rm --cached <path>`
E       assert ['scripts/__p...thon-313.pyc'] == []
FAILED tests/test_repo_hygiene.py::test_no_compiled_bytecode_is_tracked
1 failed in 0.03s
```

After `git rm --cached`, the full suite is green (188 existing + 1 new):

```
$ ruff check .
All checks passed!

$ pytest -q
189 passed in 1.51s

```

The one `validate_skill.py` warning (`body: 647 lines > 500`) is pre-existing on `master`; `SKILL.md`
is untouched by this PR.

- [x] `ruff check .` is clean

`CHANGELOG.md` had no `## [Unreleased]` section, so this adds one under `### Removed`.

Deliberately out of scope: I didn't touch the `--help` flag, which is currently rejected as an
unknown flag (`book-to-skill --help` prints `WARNING: Unknown flag '--help'` then errors on missing
input). Happy to open that as a separate issue or PR if useful — it seemed worth keeping this PR to
one change.


virgiliojr94:
Exactly the right hygiene fix — and the regression test (`git ls-files` with `-z` for non-ASCII paths, plus the defensive skips for non-git/vendored checkouts) is what makes it stick, not just a one-off removal. Committed bytecode isn't reviewable in a diff, so gating it in CI is the correct shift-left. Merging. Thanks @tjaiyen.

PR #100, opened 2026-08-04
docs: offer npx skills add install path

n0an:
## Summary (Required)
Documents installation through the cross-agent `skills` CLI (skills.sh): `npx skills add virgiliojr94/book-to-skill`. The CLI already detects the root `SKILL.md` and installs the complete skill - including `scripts/extract.py` and `tools/` - for every supported host in one command; the docs just never offered this path. `git clone` stays documented as the manual alternative.

## Type of change (Required)
- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ Feature (non-breaking change that adds capability)
- [ ] ⚡ Performance (faster / cheaper, no behavior change)
- [ ] ♻️ Refactor (no behavior change)
- [x] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling
- [ ] 💥 Breaking change (changes existing behavior or a public interface)

## What it does (Required)
- **Adds:** an `npx skills add virgiliojr94/book-to-skill` install method to the README Install section, `docs/install.md` (one-command block above the per-host clones), and `docs/index.md`. For users on any Agent Skills host who want a one-command install instead of picking the right `git clone` target per agent.

## Motivation & context (Required)
Closes n/a - no open issue; small docs gap. The install docs say the skill "follows the open Agent Skills standard, so a single install works for any compatible host", but the only agent-skill install offered is a per-host `git clone`. The `skills` CLI is the standard's one-command installer, it already works against this repo unmodified (evidence below), and the repo is already indexed at https://skills.sh/virgiliojr94/book-to-skill.

## How it works (Required for anything non-trivial)
Docs only - no code changes. The `skills` CLI resolves the repo, finds the root `SKILL.md`, and copies/symlinks the whole skill directory into each selected agent's skills folder. Because it installs the entire repo directory, the `scripts/extract.py` discovery list in `SKILL.md` resolves exactly as with a manual `git clone`.

## Evidence (Required)
- **Tests:** none needed (docs only). `ruff check .` clean, `pytest -q` 188 passed on a clean checkout.
- **Before / after:** before - no npx path documented. After - the documented command, run against this repo as-is:

```
$ npx skills add virgiliojr94/book-to-skill --skill book-to-skill -a claude-code -y --copy
◇  Source: https://github.com/virgiliojr94/book-to-skill.git
◇  Repository cloned
◇  Found 1 skill
◇  Installation complete
◇  Installed 1 skill
[...]

PR #102, opened 2026-08-04
feat(skill): optional GitHub publishing for generated skills (Step 11)

n0an:
## Summary (Required)
Adds an optional **Step 11 — Publish the generated skill to GitHub**. After a conversion finishes (and only when the Step 9.5 security scan passed), the converter offers to create a GitHub repository for the generated skill via the `gh` CLI and prints the cross-host install command (`npx skills add <repo-url> --skill <slug>`). Visibility is asked as its own closed question and `gh repo create` is **`--private` by default as a hard rule** - `--public` requires the visibility answer to *be* the bare word "public" AND the copyright gate to allow it.

## Type of change (Required)
- [x] ✨ Feature (non-breaking change that adds capability)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Adds:** `SKILL.md` Step 11 - a one-time post-report offer to publish. On acceptance the converter (1) drops a repo `README.md` into the skill folder, never overwriting an existing file; (2) runs `git init` + `gh repo create <name> --private --source . --push` in the skill folder itself, so the live install becomes the git working copy; (3) reports the repo URL and `npx skills add` command.
- **Adds:** a **hard visibility rule**. Visibility is a separate closed question asked after the publish offer ("Reply with one word: `private` or `public`"), and the gate opens only when the answer *is* `public` - substring matching is explicitly forbidden, so a sentence about the source's licence ("it's public domain", "it's publicly available") describes the material rather than the repository and still resolves to `--private`. Paraphrase, ambiguity, silence and model inference are all explicitly NOT consent: re-ask once, then fall back to `--private` and say so in the report. Stacked on top is the **copyright gate** reusing the README's Copyright & fair use policy: third-party copyrighted books stay private regardless, and access to internal material is not permission to disclose it.
- **Adds:** `tests/test_publish_visibility_gate.py` - asserts the literal `gh repo create` line keeps `--private`, that the anti-substring wording and the "public domain" carve-out survive, and a reference implementation of the gate that accepts `public` / `Public` / `public.` while rejecting `publish`, `it is public domain`, `the book is in the public domain`, `it's publicly available` and `make it public`.
- **Adds:** Update/Fold-in tie-in (a fold-in run offers to commit and push if the folder has a remote), plus a report line for the nested-repo case, where publishing from a scratch copy leaves the local folder without a remote and the push offer can therefore never fire. A "Publish your generated skill" section in `docs/usage.md` and a one-line pointer in the README Usage section.

Closes n/a - no open issue. Generated skills currently live and die on the machine that generated them; sharing one across machines or with a team means manually copying folders. The Agent Skills ecosystem already has a standard distribution channel - a git repo with a root `SKILL.md`, installable via the `skills` CLI - and generated skills already have exactly that layout. This turns publishing into a guided, policy-aware step.

Instruction-level change to `SKILL.md`, plus one test file. The generated skill's root-level `SKILL.md` layout is what the `skills` CLI detects, so the repo is installable as-is. Gating: offered only after the Step 9.5 scan passed (matching the existing "do not publish until findings are resolved" rule); `gh auth status` checked first; without `gh`, an executable fallback is printed (install/auth, or a web-created repo plus `git remote add origin && git push -u origin main`). A nested-repo guard publishes from a scratch copy when the skill folder sits inside an existing repository, so the outer project never records a gitlink without `.gitmodules`.

Rejected alternatives: defaulting visibility to a model-judged suggestion (review pointed out the *command default* must be private so a misread cannot publish); and matching the visibility answer by substring, which let a licence statement about the source open a public repo.

## Evidence (Required)
- **Tests:** `uv run --with pytest -m pytest -q` - 437 passed, 1 skipped on a clean checkout; `ruff check .` clean; `python3 tools/validate_skill.py SKILL.md` passes (pre-existing body-length soft warning, 700 lines).
- **Weight:** +1,238 tok / +53 lines against `master` measured with `tiktoken` (o200k).
- **Layout evidence:** the flow this PR documents is exactly how this repo itself is distributed - root `SKILL.md`, installable via `npx skills add virgiliojr94/book-to-skill --skill book-to-skill`. A generated skill folder matches the same detection rule, verified against a freshly generated 35-chapter skill:

```
$ ls ~/.agents/skills/clean-architecture-book
SKILL.md  chapters/  cheatsheet.md  glossary.md  patterns.md
```

- [x] `ruff check .` is clean

- All review points addressed. Latest round: the visibility gate is now a closed question matched on the bare word rather than a substring search; `gemini-extension.json` is dropped as a separate feature; the nested-repo report line closes the "publish once, never update" gap; and the visibility-gate test is included here rather than deferred, since that regression is invisible until a public repo exists that should not.
- Rebased on `c4c5e94`. The only conflict was `docs/usage.md` against the new "What people do with it" section - both kept.
- No `CHANGELOG.md` edit (git-cliff per #104).

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Two halves: a deterministic Python **extractor** (document → clean text + meta
 
 ## 🚀 Usage
 
-`/book-to-skill <path|folder|glob> [skill-name]` — plus analyze-only, generate-from-analysis, and update/fold-in modes.
+`/book-to-skill <path|folder|glob> [skill-name]` — plus analyze-only, generate-from-analysis, and update/fold-in modes. After a conversion, the converter can publish the skill to GitHub (private by default) so any host installs it with `npx skills add`.
 
 ▶️ **All modes and examples → [docs/usage.md](docs/usage.md)**
 
--- a/SKILL.md
+++ b/SKILL.md
@@ -593,18 +593,18 @@ After the Step 10 report, offer once — and only if the Step 9.5 scan passed:
 
 > "Want me to publish this skill to GitHub so any Agent Skills host can install it with `npx skills add`? (private / public / skip)"
 
-If the user declines, stop here. Requirements: the `gh` CLI, authenticated (check `gh auth status`); if unavailable, print the commands below for the user to run manually instead.
+If the user declines, stop here. Requirements: the `gh` CLI, authenticated (check `gh auth status`). If `gh` is missing or unauthenticated, offer to set it up (`brew install gh` or https://cli.github.com, then `gh auth login`) — or use the no-`gh` path: the user creates an empty repo of the chosen visibility in the GitHub web UI, then you run the `git init`/`add`/`commit` commands below followed by `git remote add origin <repo-url> && git push -u origin main`. The visibility rule below applies to the web-created repo exactly the same.
 
 **Visibility is `--private` by default — this is a hard rule, not a suggestion.** Run `gh repo create` with `--private` in every case except one: the user's reply to the question above (or a later explicit instruction) literally contains the word "public". A paraphrase, an ambiguous answer, silence, or your own inference is NOT consent — when in doubt, `--private`. A private repo can be flipped public later; a public push of book-derived content cannot be un-published.
 
-**Copyright gate — always apply before creating the repo:** chapter files are synthesized summaries, not raw text, but they still derive from the source material. Per the README's Copyright & fair use policy, skills generated from **third-party copyrighted books must stay private**; offer public only when the source is the user's own writing, internal material they hold rights to, or openly licensed content — and state which case applies.
+**Copyright gate — always apply before creating the repo:** chapter files are synthesized summaries, not raw text, but they still derive from the source material. Per the README's Copyright & fair use policy, skills generated from **third-party copyrighted books must stay private**; offer public only when the source is the user's own writing, openly licensed content, or material the user explicitly confirms they are authorized to redistribute publicly — and state which case applies. Having access to internal company material is not permission to disclose it: skills from internal docs stay **private** unless the user states they hold publication rights.
 
 If accepted:
 
 1. Add two repo files inside `$SKILLS_HOME/<skill_name>/` (never overwrite an existing file):
    - `README.md` — the skill title, a one-paragraph description ("Agent skill generated from *<Title>* by <Author> with [book-to-skill](https://github.com/virgiliojr94/book-to-skill)"), the install command from step 3 below, the file inventory, and a note that the content is synthesized summaries, not the book text.
-   - `gemini-extension.json` — `{"name": "<skill_name>", "version": "1.0.0", "description": "<description from the SKILL.md frontmatter>"}` so Gemini CLI can load the repo as an extension.
-2. Initialize the skill folder as a git repository and create the remote (default repo name `<skill_name>`; let the user override — some prefer a `<skill_name>-skill` suffix):
+   - `gemini-extension.json` — `{"name": "<skill_name>", "version": "1.0.0", "description": "<description from the SKILL.md frontmatter>", "contextFileName": "SKILL.md"}` — `contextFileName` is what makes Gemini CLI actually load the skill instructions as context, not just the manifest.
+2. Initialize the skill folder as a git repository and create the remote (default repo name `<skill_name>`; let the user override — some prefer a `<skill_name>-skill` suffix). **Nested-repo guard:** first check whether the skill folder already sits inside a git repository (`git -C "$SKILLS_HOME/<skill_name>" rev-parse --show-toplevel` — always the case for project-local roots like `.claude/skills/`). If it does, do NOT `git init` in place: the outer repository would record the folder as an embedded repo (gitlink, mode 160000) without `.gitmodules`, and fresh clones of the outer project would silently omit the skill. Instead, copy the skill folder to a scratch directory, run the commands below from the copy, and tell the user the published repo — not the project-local folder — is the remote's working copy.
 
 ```bash
 cd "$SKILLS_HOME/<skill_name>"
--- a/SKILL.md
+++ b/SKILL.md
@@ -580,9 +580,52 @@ Reload (if your agent doesn't auto-detect new skills):
   Claude Code:         restart the session
   Amp:                 restart the session
 
-Share this skill (Copilot ecosystem, optional):
-  gh skill publish $SKILLS_HOME/<skill_name>
+Share this skill (optional):
+  GitHub repo, installable on any host (Step 11):  say "publish"
+  Copilot ecosystem:  gh skill publish $SKILLS_HOME/<skill_name>
 ```
+
+---
+
+## Step 11 — Publish the generated skill to GitHub (optional)
+
+After the Step 10 report, offer once — and only if the Step 9.5 scan passed:
+
+> "Want me to publish this skill to GitHub so any Agent Skills host can install it with `npx skills add`? (private / public / skip)"
+
+If the user declines, stop here. Requirements: the `gh` CLI, authenticated (check `gh auth status`); if unavailable, print the commands below for the user to run manually instead.
+
+**Visibility is `--private` by default — this is a hard rule, not a suggestion.** Run `gh repo create` with `--private` in every case except one: the user's reply to the question above (or a later explicit instruction) literally contains the word "public". A paraphrase, an ambiguous answer, silence, or your own inference is NOT consent — when in doubt, `--private`. A private repo can be flipped public later; a public push of book-derived content cannot be un-published.
+
+**Copyright gate — always apply before creating the repo:** chapter files are synthesized summaries, not raw text, but they still derive from the source material. Per the README's Copyright & fair use policy, skills generated from **third-party copyrighted books must stay private**; offer public only when the source is the user's own writing, internal material they hold rights to, or openly licensed content — and state which case applies.
+
+If accepted:
+
+1. Add two repo files inside `$SKILLS_HOME/<skill_name>/` (never overwrite an existing file):
+   - `README.md` — the skill title, a one-paragraph description ("Agent skill generated from *<Title>* by <Author> with [book-to-skill](https://github.com/virgiliojr94/book-to-skill)"), the install command from step 3 below, the file inventory, and a note that the content is synthesized summaries, not the book text.
+   - `gemini-extension.json` — `{"name": "<skill_name>", "version": "1.0.0", "description": "<description from the SKILL.md frontmatter>"}` so Gemini CLI can load the repo as an extension.
+2. Initialize the skill folder as a git repository and create the remote (default repo name `<skill_name>`; let the user override — some prefer a `<skill_name>-skill` suffix):
+
+```bash
+cd "$SKILLS_HOME/<skill_name>"
+git init -b main
+git add -A
+git commit -m "Add <skill_name> skill"
+gh repo create <repo_name> --private --source . --push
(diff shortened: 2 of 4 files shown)


virgiliojr94:
This is the right shape for a publish step — opt-in, never automatic, private-by-default, and it reuses the existing Copyright & fair-use policy for the public/private gate. That's exactly the posture an outward-facing action needs. Nice.

Two things before merge:
1. Since the public/private choice is model-followed prose, please make the `gh repo create` command default to `--private` unless the user explicitly typed "public" — so a misread can never create a public repo from a copyrighted, book-derived skill. (The draft mostly does this; I just want the *command default* to be private, not the model's discretion.)
2. It conflicts with #100 on `SKILL.md`/`README`/`CHANGELOG`, so we'll sequence the two — whichever lands second rebases.

Solid contribution. Fix (1) and we're good.

virgiliojr94:
To unblock: (1) make the `gh repo create` command default to `--private` unless the user explicitly chose public (so a misread can't publish a copyrighted, book-derived skill); (2) from #104, the changelog is now git-cliff-generated — drop your `CHANGELOG.md` edit on rebase. Title is already conventional. The opt-in/private-default publish model is otherwise good to go.

n0an:
@virgiliojr94 Both points addressed - could you take another look?

1. **`--private` is now the literal command default, not model discretion.** Step 11 states a hard visibility rule: run `gh repo create --private` in every case except when the user's reply literally contains the word "public" - paraphrase, ambiguity, silence, or model inference is explicitly NOT consent. The command block itself carries the same restriction, and the copyright gate stacks on top (third-party books stay private regardless of what the user typed).
2. **Changelog edit dropped** per the git-cliff policy from #104, and the branch is **rebased on latest `master`** - the usage docs moved into `docs/usage.md` to match the README restructure.

Validation on the rebased branch: 270 passed / 1 skipped, ruff clean, `validate_skill.py` passes.

n0an:
Addressed all four Codex review points in `b19928b`:

- **P1 (internal-source rights):** the copyright gate no longer treats "internal material they hold rights to" as public-eligible - access is not permission to disclose. Internal docs stay **private** unless the user explicitly states they hold publication rights; public remains possible only for own writing, openly licensed content, or explicitly authorized material.
- **P2 (nested repos):** Step 11 now has a nested-repo guard - if the skill folder sits inside an existing git repository (always true for project-local roots), it is published from a scratch copy instead of `git init` in place, avoiding an embedded-repo gitlink without `.gitmodules` in the outer project.
- **P2 (gh fallback):** the no-`gh` path is now actually executable - install/auth instructions, or a web-created repo plus `git remote add origin && git push -u origin main`, with the same visibility rule applying.
- **P2 (Gemini context):** `gemini-extension.json` now includes `"contextFileName": "SKILL.md"` so Gemini CLI loads the skill content, not just the manifest.

virgiliojr94:
Re-reviewed the rebased branch. Both of my earlier asks are done properly. I also went back and measured the objection I was about to raise, and most of it did not survive — so this review is shorter than the one I had drafted, and points somewhere else.

**Verified rather than taken on trust:**

- `--private` is a literal command default with an explicit "paraphrase / silence / inference is NOT consent" rule, and the copyright gate stacks on top.
- The internal-rights fix is right: access is not permission to disclose.
- The nested-repo guard reasoning holds — `git init` inside an outer repo records a gitlink at mode 160000 with no `.gitmodules`, and clones of the outer project would silently omit the skill.
- The install command works. Against a clean `HOME`: `npx skills add https://github.com/virgiliojr94/book-to-skill --skill book-to-skill` installs; `--skill does-not-exist` fails with `No matching skills found`. `-s, --skill` is real, and Step 8 writes `name: <skill_name>` into the generated frontmatter, so the reported command matches what the CLI expects.
- 270 passed, ruff clean, `validate_skill.py` passes.

**On the weight of the diff — I was going to block on this and I was wrong.**

Measured with `tiktoken` (o200k), not the project's word-based proxy:

```
SKILL.md    master 7,617 tok / 647 lines
            PR     8,675 tok / 690 lines
            delta  +1,058 tok (+13.9%)
```

Against `docs/performance.md`, a full conversion is 155K–391K input tokens (~\$1/book). So the added weight is **0.3–0.7% of a run, about \$0.003**. And per Anthropic's Skills guidance, the body is not resident at all times — "only the metadata (name and description) from all Skills is pre-loaded… Claude reads SKILL.md only when the Skill becomes relevant". It loads exactly when a conversion is about to spend six figures of tokens. A cost objection does not survive that denominator, so I withdraw it.

What does survive is structural, and it is a suggestion, not a gate. The 500-line threshold in `validate_skill.py` is not something we invented — it is the documented guidance ("Keep SKILL.md body under 500 lines for optimal performance", "Split content into separate files when approaching this limit"), and Pattern 3 in that same guide, "Conditional details: show basic content, link to advanced content", describes this step exactly. So `references/publishing.md` with three lines left in Step 11 is the shape I'd like eventually. But `master` is already 147 lines over that line with conditional steps inline (2.6, 9.5), and holding your 43 lines to a standard the file has never met would be me charging you for my debt. Land it inline; the split is my job, across the whole file, in its own PR.

For the record on my own remedy: I checked whether moving content out would hide it from tooling. `validate_skill.py` reads a single file and does not follow references — but all it checks is frontmatter and line count, so nothing load-bearing would be lost. Worth knowing before anyone does the split.

---

**One real change before merge, and it is in the safety rule itself.**

The visibility gate fires on the reply "(or a later explicit instruction)" *literally containing the word "public"*. That token is not as rare as it looks in this project:

```
reply                        gate opens public repo?
"publish"                    no
"publish it"                 no
"make it public"             yes   <- intended
"it is public domain"        yes   <- NOT a visibility decision
"publicly"                   yes
```

"Public domain" is ordinary vocabulary here: Moby-Dick is one of our own benchmark books, and the README's policy invites exactly this sentence by naming "openly-licensed material" as the shareable case. So a user explaining *the source's licence* — the one topic the copyright gate is actively asking them about — can satisfy the *visibility* gate without ever choosing to publish publicly. Two different questions, one keyword.

(I also chased the reverse — whether "publish" contains "public" and self-triggers. It does not: `"public" in "publish"` is `False`. Mentioning it so nobody re-derives that scare later.)

Cheapest fix that keeps your design: make the visibility answer a closed one rather than a substring search — ask visibility as its own question and require the reply to *be* `public` (bare word), not merely contain it, and state that a sentence about the source's licence is never a visibility answer. Everything else in the rule stays.

**Still take, unchanged:** drop `gemini-extension.json`. This PR is "publish to GitHub"; a host-specific manifest for the one agent this project doesn't otherwise support is a separate feature, and the PR itself says a root `SKILL.md` is all the `skills` CLI needs.

**Two one-liners, non-blocking:**

- Project-local installs publish once and can never update: you correctly publish from a scratch copy, but Update/Fold-in only offers a push when "the skill folder is a git repository with a remote", which that folder never becomes. One sentence in the Step 11 report closes it.
- Nothing in `tests/` asserts anything about `SKILL.md`'s content. The visibility rule is the one behaviour here where a regression is invisible until a public repo exists that shouldn't. A tiny test asserting the gate's wording (and that the Step 10 trigger word can't satisfy it) would be cheap insurance — happy to take it in a follow-up rather than hold this PR.

Fix the visibility keyword and drop the Gemini manifest, and this goes in.

n0an:
Both blockers addressed in `b855b5e`, plus the two one-liners.

**1. The visibility gate is no longer a substring search.** You were right that "public" is not a rare token in this project, and the fix is the closed question you proposed:

- The publish offer is now `(yes / skip)` - it no longer carries the visibility choice at all.
- Visibility is asked on its own: *"Private or public repository? Reply with one word: `private` or `public`."*
- The gate opens only when the answer **is** `public`. Whitespace, casing and trailing punctuation are normalized; nothing else is. Substring matching is forbidden in the rule text, not merely discouraged.
- A licence statement is named as a non-answer: "it's public domain", "the book is public domain", "it's publicly available" all describe the material rather than the repository, and all resolve to `--private`.
- Anything else re-asks once, then falls back to `--private` and says so in the report.

**2. `gemini-extension.json` dropped.** Agreed, and for your reason: a root `SKILL.md` is all the `skills` CLI needs, so the manifest was a second feature riding along on a publish step. Step 11 now writes only the repo `README.md`. This does remove the file that the Codex `contextFileName` fix applied to; if Gemini support lands later it should arrive with its own PR and its own justification.

**3. Project-local publishes now say they cannot update.** Step 11's report gains a conditional line for the scratch-copy path, since that folder never gains a remote and the Update/Fold-in push offer can therefore never fire for it:

```
⚠️  Published from a copy: <skill folder> sits inside another git repository, so it has
    no remote of its own. To publish a later update, re-run Step 11, or clone
    https://github.com/<owner>/<repo_name> and fold new material into the clone.
```

**4. I took the test rather than deferring it.** `tests/test_publish_visibility_gate.py`. Your framing is what convinced me it belongs here rather than in a follow-up: this is the one behaviour where the regression is invisible until a public repo exists that should not.

It asserts two independent things. First, that the prose still carries the rule: the literal `gh repo create` line keeps `--private` and never hardcodes `--public`, "substring matching is forbidden" survives, and the "public domain" carve-out is still named. Second, a reference implementation of the gate, checked against a table:

| reply | gate |
|---|---|
| `public`, `Public`, `  public  `, `public.` | opens |
| `publish`, `publish it` | closed (the Step 10 trigger word cannot satisfy it) |
| `it is public domain`, `the book is in the public domain`, `it's publicly available`, `public domain` | closed |
| `make it public`, `public is fine i guess`, `yes`, `private`, empty | closed |

The second half executes nothing from `SKILL.md` - nothing in this project does. It pins the intended semantics next to the prose, so a future edit that loosens the wording fails the first half while the second still shows what the wording was supposed to mean. If you would rather it lived somewhere else, or asserted less, say so and I will move it.

**Rebased** on `c4c5e94`. One conflict, in `docs/usage.md`, against your new "What people do with it" section - both kept, publish section placed after it.

**On the weight, since you measured it rather than guessing:** the branch is now +1,238 tok / +53 lines against `master` (o200k), up from the +1,058 / +43 you measured. The visibility rule grew by the separate question and the licence carve-out; dropping the Gemini bullet gave a little back. Flagging the delta rather than letting you rediscover it - the conclusion you reached about the denominator is unchanged, but you should not have to re-measure to find that out.

**Validation on the rebased branch:** 437 passed / 1 skipped, `ruff check .` clean, `python3 tools/validate_skill.py SKILL.md` passes with the pre-existing body-length soft warning (700 lines).

virgiliojr94:
Taking it. The visibility gate is now a closed question with its own answer space, which is the shape I wanted and better than what I asked for — the offer and the visibility choice are separate prompts, so the Step 10 trigger word cannot reach the gate at all.

The test is the part I want to single out. I suggested it as a follow-up and you took it into the PR with the right argument: this is the one behaviour whose regression is invisible until a public repo exists that should not. Asserting the prose *and* pinning a reference implementation of the gate against a table is a good answer to "you cannot test prose" — the first half fails if someone loosens the wording, the second half documents what the wording was supposed to mean. `public domain`, `publicly available`, `publish it` all closed; `public` alone opens. That table is the spec.

The nested-repo notice reads right too: the user is told the published repo has no local remote and what to do at update time, instead of discovering it when a fold-in has nowhere to push.

On the weight: thank you for re-measuring and flagging the delta rather than letting me find it. +1,238 tokens against a run that costs 155K–391K is 0.3–0.8%, and the body of `SKILL.md` is not resident until the skill is triggered — which is when a conversion is about to happen anyway. The conclusion stands: not a reason to block, and the 700-line body is my debt to pay across the whole file, not yours to pay in this PR.

Merging.

PR #92, opened 2026-08-01
fix(extractor): detect Markdown-prefixed chapter headings (#91)

Stamina9:
## Summary (Required)

  Fixes chapter detection for books whose chapters are headed with a Markdown
  prefix (`## Chapter 1`) — the exact form Docling emits in `--mode technical`.
  `_chapter_number()` now strips a leading `#`/`=` heading marker and matches the
  heading underneath, so `chapters_detected` no longer falls through to the
  structural fallback and inflates. Non-headings stay rejected.

  ## Type of change (Required)

  - [x] 🐛 Bug fix (non-breaking change that fixes an issue)

  ## What it does (Required)

  - **Fixes:** `_chapter_number()` returned `None` for `## Chapter 1`,
    `## Capítulo 5`, `== Section`, so `--mode technical` books (Docling emits
    headings as Markdown) fell through to `_structural_chapter_count()`, which
    inflated `chapters_detected` — a 2,718-page volume reported 375 "chapters"
    against 35 real ones (#91).
  - **Improves:** the two detection paths (plain-text vs Markdown) now agree: the
    #91 repro reports 35 chapters for both, and the heading sample is filled
    instead of empty (an empty sample is the reliable tell that the structural
    fallback was used).
  - **Adds:** regression tests for Latin / Roman / AsciiDoc prefixed headings,
    non-heading rejection, and CJK behavior unchanged.

  ## Motivation & context (Required)

  Closes #91

  ## How it works (Required for anything non-trivial)

  `_chapter_number()` runs a second pass: if the first match fails and the line
  starts with a Markdown/AsciiDoc marker (`#{1,6}` or `={1,6}` + space), it
  strips the prefix and retries the same matchers. Stripping once at the top of
  `_chapter_number()` covers every language variant in one place (the fix #91
  suggested); the CJK/Thai/Korean matchers that already tolerated the prefix
  inline are untouched because their regexes are unchanged and the stripped
  string is identical to the plain-text form.

  ## Evidence (Required)

  - **Tests:** added `TestMarkdownPrefixedLatinChapters` (7 tests) asserting
    `## Chapter 1` → 1, `== Chapter 1` → 1, `## I. Loomings` → 1, the #91 repro
    now reports 35 (matching the plain-text path), and `## Some Section` /
    `## 5 Setup` stay `None`. Full suite: 271 passed on Python 3.12 (rebase
    target included).
  - **Before / after:**

  ▎ ▎ ▎ from book_to_skill.utils import _chapter_number
[...]

PR #107, opened 2026-08-04
fix(html): emit block boundaries at end tags in the stdlib parser

Hotragn:
## Summary

The stdlib HTML parser only emitted a line break on a block element's *opening* tag, and its tag list omitted every table and definition-list element. Text from adjacent blocks therefore ran together — an entire table row collapsed into one token run (`200OKRequest succeeded`). Boundaries are now emitted on close as well, and the tag set covers the block elements it was missing.

## Type of change

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does

- **Fixes:** `_HTMLTextExtractor` appended `"\n"` in `handle_starttag` only, for the fixed list `("p", "br", "h1"…"h6", "li", "div")`. `handle_endtag` appended nothing, and `td`, `th`, `tr`, `table`, `dt`, `dd`, `section`, `article`, `blockquote`, `pre` were absent. So any two text runs separated only by a closing tag or an unlisted tag were concatenated with no separator.
- **Improves:** on a 5-row HTTP status-code reference table, 7 of 7 merged-token probes were present before and **0** after; the whitespace-token count goes from 21 to the correct **31**. Four heading layouts go from **0** detected chapters to the correct **3**.

## Motivation & context

Closes n/a — found while auditing the dependency-free extraction path; no existing issue.

This is the fallback used whenever `beautifulsoup4` is absent, which is the path CI's own `smoke (dependency-free extraction)` job exercises, and `parsers/epub.py` imports `_HTMLTextExtractor` so the stdlib EPUB extractor inherits the same defect.

Two distinct harms:

1. **Table content is corrupted** (unconditional). Reference and technical books are full of tables; every row lost its cell separation, so terms merged into non-words that then pollute the Step 8 glossary and shift `words` / `estimated_tokens` in `metadata.json`.
2. **Chapter detection silently fails** in layouts where the affected element is the only place a chapter number appears. `_EXPLICIT_CHAPTER` requires `\b` after the number; in `Chapter 1Introduction` both `1` and `I` are word characters, so there is no boundary, the regex does not match, and the heading is not counted. Extraction still reports success with `chapters_detected: 0`.

In scope-honesty: a book whose chapter bodies happen to be `<h2>Chapter 1</h2><section><p>…` still detected fine before this PR, because the `<p>` opening tag supplied a break. The chapter-detection failure needs the heading to be followed directly by text, or by an element that was not in the old list. The **table-cell merging is unconditional** and is the primary defect here.

## How it works

- `BLOCK_TAGS` — block-level elements. A boundary is emitted when they open **and** when they close.
- `CELL_TAGS` (`td`, `th`) — joined with `"\t"` rather than `"\n"`, so a table row stays on one line. This matches the convention `parsers/docx.py` already uses for tab-joined table rows (#61), and it is what keeps a table-formatted ToC row parseable as a heading: `Chapter 1\tIntroduction\t1` satisfies `_HEADING_TAIL` (`^\s+[A-ZÀ-Þ0-9…]`), whereas `Chapter 1\nIntroduction` would split the row across lines.
- Boundaries are **deferred** (`self._pending`) until the next non-blank text run, rather than appended immediately. This gives three properties for free: nested blocks (`<div><div><p>`) collapse to one separator instead of a run of blank lines; there is no leading blank line; and whitespace-only layout indentation between tags is dropped rather than emitted as trailing spaces.
- `"\n"` outranks `"\t"` in `_mark()`, so a `<td>` opening straight after a `<tr>` cannot downgrade the row boundary to a cell boundary.
- **Inline elements are deliberately untouched.** `hyper<span>text</span>` must stay one word, and `<b>bold</b> <i>italic</i>` must keep its space. Both are asserted.

Rejected alternative: emitting `"\n"` between every string, which is what bs4's `get_text(separator="\n")` does. It fixes the merging but puts every table cell on its own line, so a table-formatted ToC row can no longer be recognised as a single heading — strictly worse for this project's purpose.

Also removed the unused `_current_skip` attribute.

## Evidence

**Before / after** — a reference table, the shape technical books use constantly:

```
############ BEFORE ############
  'HTTP status codes'
  'CodeNameWhen to use'
  '200OKRequest succeeded'
  '301Moved PermanentlyResource has a new URL'
  '404Not FoundNo such resource'
  '503Service UnavailableServer overloaded'

merged-token probes present: 7/7 -> ['OKRequest', 'PermanentlyResource', 'FoundNo',
                                     'UnavailableServer', 'CodeName', '200OK', '301Moved']
distinct whitespace-separated tokens: 21

############ AFTER ############
  'HTTP status codes'
  'Code\tName\tWhen to use'
  '200\tOK\tRequest succeeded'
  '301\tMoved Permanently\tResource has a new URL'
  '404\tNot Found\tNo such resource'
  '503\tService Unavailable\tServer overloaded'

merged-token probes present: 0/7 -> []
distinct whitespace-separated tokens: 31
```

**Chapter detection**, `detect_structure()` on four layouts that occur in converted ebooks:

| layout | before | after | bs4 path |
|---|---:|---:|---:|
| table-formatted ToC | **0** | 3 | 3 |
| `<h2>…</h2><section>` | **0** | 3 | 3 |
| `<h2>…</h2>` + bare text | **0** | 3 | 3 |
| `<dl>/<dt>/<dd>` ToC | **0** | 3 | 3 |

The bs4 column shows this was also a divergence between the two paths, not just an absolute bug.

**Tests** — new `tests/test_html_block_boundaries.py`, 16 cases in five groups:

- `TestBlockBoundaryChapterDetection` — the four layouts above (parametrized), plus a row-stays-on-one-line assertion and a direct `"Chapter 1Reliable" not in text` check.
- `TestInlineTextUnchanged` — inline whitespace preserved, `hyper<span>text</span>` stays one word, an anchor mid-sentence stays inline. These are the regression guards for the fix itself.
- `TestSeparatorHygiene` — no leading separator, nested blocks collapse, layout whitespace dropped, `<br/>` still breaks, skip-tag content still excluded.
- `TestConvergenceWithBeautifulSoup` — the fallback and the bs4 path agree on chapter count (`importorskip`, so it is skipped when bs4 is absent).
- `TestEpubStdlibPath` — builds a real EPUB and asserts the fix reaches `extract_with_zipfile`.

RED against unmodified `master`: **16 failed**. GREEN with the fix, full suite:

```
$ python -m pytest tests/ -q
209 passed in 1.96s

$ ruff check .
All checks passed!
```

New file rather than appended to `test_book_to_skill.py`, to stay conflict-free against the other open PRs touching that file.

## Screenshots / output

The before/after text dumps above are the output — a terminal diff shows the merged-vs-separated cells more precisely than an image would.

## Checklist

- [x] `ruff check .` is clean

## Notes for reviewers

The one judgement call worth a second opinion is `"\t"` for cells vs `"\n"`. I chose `"\t"` to keep table ToC rows parseable and to match the DOCX fallback, but it does mean the HTML and bs4 paths still produce different *text* (same chapter count). If you would rather the two paths produce byte-identical output, say so and I will switch cells to `"\n"` and drop the `test_table_row_stays_on_one_line` assertion.

`BLOCK_TAGS` includes `br` even though it is void — `handle_startendtag` dispatches to both handlers, and `_mark` is idempotent, so `<br/>` and `<br>` both yield exactly one break.

### DIFF
--- a/book_to_skill/parsers/html.py
+++ b/book_to_skill/parsers/html.py
@@ -10,25 +10,73 @@ class _HTMLTextExtractor(html.parser.HTMLParser):
 
     SKIP_TAGS = {"script", "style", "head"}
 
+    # Block-level elements. A boundary is emitted both when they open and when
+    # they CLOSE — closing matters, because without it the text of two adjacent
+    # blocks concatenates ("<h2>Chapter 1</h2>Intro" -> "Chapter 1Intro"), which
+    # destroys chapter detection: _EXPLICIT_CHAPTER requires a word boundary
+    # after the number, and "1I" has none.
+    BLOCK_TAGS = frozenset({
+        "address", "article", "aside", "blockquote", "br", "dd", "details",
+        "div", "dl", "dt", "fieldset", "figcaption", "figure", "footer",
+        "form", "h1", "h2", "h3", "h4", "h5", "h6", "header", "hgroup", "hr",
+        "li", "main", "nav", "ol", "p", "pre", "section", "table", "tbody",
+        "tfoot", "thead", "tr", "ul",
+    })
+    # Table cells are separated by a tab rather than a newline so a row stays on
+    # one line — the same convention the stdlib DOCX fallback already uses for
+    # tab-joined table rows, and what keeps a table-formatted table of contents
+    # ("Chapter 1 | Introduction | 1") parseable as a single heading line.
+    CELL_TAGS = frozenset({"td", "th"})
+
     def __init__(self):
         super().__init__()
         self._parts: list[str] = []
         self._skip_depth = 0
-        self._current_skip: str | None = None
+        # Strongest boundary awaiting the next non-blank text run. Deferring it
+        # (instead of appending immediately) means nested blocks such as
+        # "<div><p>x" collapse to one separator rather than a run of blank lines.
+        self._pending = ""
+
+    def _mark(self, separator: str) -> None:
+        # "\n" outranks "\t": a row/block boundary must not be downgraded to a
+        # cell boundary by a <td> that opens straight after a <tr>.
+        if separator == "\n" or not self._pending:
+            self._pending = separator
 
     def handle_starttag(self, tag, attrs):
         if tag in self.SKIP_TAGS:
             self._skip_depth += 1
-        if tag in ("p", "br", "h1", "h2", "h3", "h4", "h5", "h6", "li", "div"):
-            self._parts.append("\n")
+        if tag in self.BLOCK_TAGS:
+            self._mark("\n")
+        elif tag in self.CELL_TAGS:
+            self._mark("\t")
 
     def handle_endtag(self, tag):
-        if tag in self.SKIP_TAGS and self._skip_depth:
-            self._skip_depth -= 1
+        if tag in self.SKIP_TAGS:
+            if self._skip_depth:
+                self._skip_depth -= 1
+            return
+        if tag in self.BLOCK_TAGS:
+            self._mark("\n")
+        elif tag in self.CELL_TAGS:
+            self._mark("\t")
 
     def handle_data(self, data):
-        if not self._skip_depth:
-            self._parts.append(data)
+        if self._skip_depth:
+            return
+        if self._pending:
+            if not data.strip():
+                # Whitespace-only text between tags is layout indentation. It
+                # cannot satisfy a pending boundary, and emitting it before the
+                # boundary would just add trailing spaces — drop it and keep
+                # waiting for real content.
+                return
+            # Suppress a leading separator so the output does not start with a
+            # blank line.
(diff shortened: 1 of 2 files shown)


virgiliojr94:
Excellent fix — the heading-glue bug (`<h2>Chapter 1</h2>Intro` → "Chapter 1Intro" → 0 chapters) is real, same class as the SOURCE-banner (#83) and markdown-prefix (#92) cases. The deferred-separator approach is right: fixes adjacent-block runs *and* dedupes nested blocks, while leaving inline tags (`<b>`/`<span>`/`<a>`) boundary-free so words don't split — and your tests prove exactly that, incl. the bs4-convergence and EPUB-path cases. Merging. Thanks @Hotragn.

PR #109, opened 2026-08-04
fix(rtf): skip control destination groups in the dependency-free fallback

Hotragn:
## Summary

The dependency-free RTF fallback stripped control *words* and then removed all braces, but never skipped RTF **destination groups**. Everything inside `\fonttbl`, `\colortbl`, `\stylesheet`, `\*\generator` and `\info` therefore survived as literal text — font names, style names, the generator string, and the document title and author all landed in the extracted book text. All seven leak probes leaked before; none leak now.

## Type of change

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does

- **Fixes:** `strip_rtf_fallback` had no concept of a destination group. `re.sub(r"\\[a-zA-Z]+-?\d* ?", "", raw)` removed `\fonttbl` and `\f0\fnil\fcharset0`, then `.replace("{", "")` removed the braces — leaving `Calibri;` sitting in the prose as though the book had said it.
- **Improves:** 7 of 7 leak probes eliminated. Also fixes escaped-brace mangling: `A set \{a, b\}` extracted as `A set \a, b\` and now extracts as `A set {a, b}`.

## Motivation & context

Closes n/a — found while auditing the dependency-free extraction paths; no existing issue.

Four concrete consequences of the leak:

1. **Glossary pollution.** Font and style names (`Calibri`, `Helvetica Neue`, `heading 1`, `Normal`) are exactly the kind of short capitalised tokens Step 8 picks up as candidate terms.
2. **Metadata exposure.** `\info` carries `\title` and `\author`, and can also carry `\doccomm` comments and revision data. That is copied into a file the user may publish as a skill — the README's own workflow.
3. **Skewed cost pre-flight.** The residue inflates `words` and `estimated_tokens` in `metadata.json`.
4. **Path divergence.** `striprtf` handles destination groups correctly, so a user with the optional dependency installed got clean text and a user without it got junk — the same fallback-vs-library gap as #52, #60, #61 and #64.

## How it works

`_strip_destination_groups` is a brace-depth scanner that runs *before* the existing control-word cleanup, so a skipped group's contents never reach the sweeps that would strip its markup and leave the names behind.

- **Named non-content destinations** are listed explicitly in `_SKIP_DESTINATIONS` with a comment showing what each one used to leak.
- **`\*` ignorable destinations are skipped generically.** Per the RTF spec a reader that does not understand a `\*` destination must skip the whole group, so this covers `\*\generator`, `\*\bkmkstart` and any vendor extension without having to enumerate them. It also gives correct field handling for free: `{\field{\*\fldinst HYPERLINK …}{\fldrslt visible text}}` keeps the result and drops the instruction.
- **Escaped literals** `\{`, `\}`, `\\` are not treated as delimiters. The scanner has to special-case them anyway to count depth correctly, so the pre-existing mangling (a stray `\` where the book said `{a, b}`) is fixed in the same pass via placeholder parking around the brace sweep. Flagging this as a small adjacent fix rather than hiding it.
- **Malformed input cannot truncate.** If a destination group is never closed, the scan would drop everything after the unclosed brace — possibly the entire book. In that case it returns the unscanned text instead, so this change cannot lose a document that previously extracted. Verified by `test_unterminated_skipped_group_falls_back`.

Deliberately **not** skipped: `\footnote` (real content), and `\upr` — it wraps duplicate ANSI/Unicode representations of the same run, so skipping it would lose text.

## Evidence

**Before / after** on a realistic RTF header (font table, colour table, stylesheet, generator, `\info`):

```
BEFORE
'\nCalibri;Helvetica Neue;\n;;;\nNormal;heading 1;\n\\*Riched20 10.0.19041;
 \nSecret DraftJane Roe\n\nChapter 1\n\nReal body text here.\n'

  Calibri          LEAKED
  Helvetica Neue   LEAKED
  Normal           LEAKED
  heading 1        LEAKED
  Riched20         LEAKED
  Secret Draft     LEAKED
  Jane Roe         LEAKED

AFTER
'\n\n\n\n\n\n\nChapter 1\n\nReal body text here.\n'

  Calibri          ok
  Helvetica Neue   ok
  Normal           ok
  heading 1        ok
  Riched20         ok
  Secret Draft     ok
  Jane Roe         ok

content kept: True
```

**Edge cases**, all passing:

```
escaped braces are literal text                  PASS  -> 'A set {a, b} of items.'
field keeps result, drops instruction            PASS  -> 'See chapter 4 now.'
nested skip inside skip                          PASS  -> 'Body.'
content group after skipped group                PASS  -> 'Bold text\n tail'
unicode escape survives outside skipped groups   PASS  -> '“Hi”'
pict binary payload dropped                      PASS  -> 'Caption here.'
escaped backslash preserved                      PASS  -> 'Path C:\temp\out'

unterminated skipped group -> 'Arial; Body text never closed'   (falls back, not truncated)
```

**Tests** — new `tests/test_rtf_destination_groups.py`, 20 cases in five groups:

- `TestDestinationGroupsDropped` — the seven probes (parametrized), body text survives, the heading is on its own line, `\info` not exposed, `\pict` hex payload dropped, nested-skip, and a content group after a skipped group still kept.
- `TestIgnorableDestinations` — an unknown `\*` vendor destination is skipped; the field instruction/result split.
- `TestEscapedLiteralsSurvive` — `\{`/`\}`/`\\`.
- `TestExistingBehaviourPreserved` — regression net for the `\uN` decoding from #52, and `\par`/`\tab`.
- `TestMalformedInputIsNotTruncated` — the unterminated-group fallback.

RED against unmodified `master`: **17 failed, 3 passed**. GREEN with the fix:

```
$ python -m pytest tests/ -q
213 passed in 2.32s

$ ruff check .
All checks passed!
```

## Screenshots / output

The before/after text dumps above are the output; a terminal diff shows the leaked font/metadata strings more precisely than an image would.

## Checklist

- [x] `ruff check .` is clean

## Notes for reviewers

The escaped-literal fix is technically a second bug in the same function. I folded it in because the depth scanner must handle `\{`/`\}` correctly regardless, so fixing the downstream mangling was two lines rather than a separate PR that would conflict with this one. Happy to split it out if you would rather keep the diff to destinations only.

`_SKIP_DESTINATIONS` is deliberately a conservative allowlist-of-exclusions rather than "skip every unrecognised destination". A more aggressive rule would drop unknown *non-`\*`* destinations too, but the `\*` prefix exists precisely to mark "safe to skip", so anything without it may well be content and I would rather under-skip than eat a chapter.

Three leftover blank lines appear where the header groups were (visible in the AFTER dump). Harmless — `detect_structure` is line-based and blank lines do not affect it — but if you would prefer the fallback to collapse runs of blank lines, that is an easy follow-up.

### DIFF
--- a/book_to_skill/parsers/rtf.py
+++ b/book_to_skill/parsers/rtf.py
@@ -19,13 +19,112 @@ def _rtf_unicode_repl(match: re.Match) -> str:
     return chr(cp)
 
 
+# RTF groups whose contents are metadata or formatting tables rather than
+# document text. Stripping only the control words inside them (what the cleanup
+# below does) leaves the residue behind: font and style *names*, the generator
+# string, and the \info title/author all end up in the extracted book text.
+_SKIP_DESTINATIONS = frozenset({
+    "fonttbl",            # {\fonttbl{\f0\fnil Calibri;}}   -> "Calibri;"
+    "colortbl",           # {\colortbl;\red255...;}          -> ";;;"
+    "stylesheet",         # {\stylesheet{\s0 Normal;}}       -> "Normal;"
+    "info",               # {\info{\title X}{\author Y}}     -> "XY"
+    "listtable", "listoverridetable", "revtbl", "rsidtbl",
+    "latentstyles", "datastore", "themedata", "colorschememapping",
+    "filetbl", "xmlnstbl", "pgptbl", "protusertbl", "userprops",
+    "docvar",
+    "pict", "objdata",    # binary image / OLE payloads as hex text
+    "bkmkstart", "bkmkend",
+})
+
+# The first control word of a group, allowing the "\*" ignorable-destination
+# prefix: "{\fonttbl", "{\*\generator", "{\*\bkmkstart".
+_GROUP_DESTINATION = re.compile(r"\\\*?\\?([a-zA-Z]+)")
+
+
+def _strip_destination_groups(raw: str) -> str:
+    """Remove RTF groups that hold no document text.
+
+    Tracks brace depth so a whole group is dropped, not just its control words.
+    Per the RTF spec a reader that does not understand a ``\\*`` destination must
+    skip the entire group, which also handles ``\\*\\generator`` and any vendor
+    extension without naming it. Escaped ``\\{`` / ``\\}`` / ``\\\\`` are not
+    treated as delimiters.
+
+    A useful side effect: for a field, ``{\\field{\\*\\fldinst HYPERLINK ...}
+    {\\fldrslt visible text}}`` keeps the result and drops the instruction.
+    """
+    out: list[str] = []
+    index = 0
+    depth = 0
+    skip_at_depth = 0  # non-zero while inside a skipped group
+    length = len(raw)
+
+    while index < length:
+        char = raw[index]
+
+        # Escaped literal: "\{", "\}", "\\" are text, never group delimiters.
+        if char == "\\" and index + 1 < length and raw[index + 1] in "{}\\":
+            if not skip_at_depth:
+                out.append(raw[index:index + 2])
+            index += 2
+            continue
+
+        if char == "{":
+            depth += 1
+            if not skip_at_depth:
+                match = _GROUP_DESTINATION.match(raw, index + 1)
+                ignorable = raw.startswith("{\\*", index)
+                if ignorable or (match and match.group(1) in _SKIP_DESTINATIONS):
+                    skip_at_depth = depth
+                else:
+                    out.append(char)
+            index += 1
+            continue
+
+        if char == "}":
+            if skip_at_depth and depth == skip_at_depth:
+                skip_at_depth = 0
+            elif not skip_at_depth:
+                out.append(char)
+            depth -= 1
+            index += 1
+            continue
+
+        if not skip_at_depth:
+            out.append(char)
+        index += 1
+
+    if skip_at_depth:
+        # Unterminated destination group: the file is malformed and everything
+        # after the unclosed brace was just dropped, which could be the whole
+        # book. Leaking some metadata residue is the lesser evil, so fall back
+        # to the unscanned text rather than returning a truncated document.
+        return raw
+
+    return "".join(out)
+
+
 def strip_rtf_fallback(raw: str) -> str:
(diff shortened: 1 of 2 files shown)


virgiliojr94:
Nice work — the brace-depth scanner is the right approach (spec-compliant "skip the whole \\* group"), and dropping `\fonttbl`/`\info` residue before the control-word sweep fixes real leakage into the glossary and `metadata.json`. Tests cover the branches that matter (nesting, ignorables, escaped braces, malformed-input fallback). Tiny non-blocker: `bkmkstart`/`bkmkend` in the skip set are redundant (bookmarks are always `\*`-prefixed, already caught) — harmless, leave or drop. Merging. Thanks @Hotragn.

Hotragn:
Thanks for the quick review on all four.

On the `bkmkstart` / `bkmkend` non-blocker — you're right for spec-conforming RTF, but I checked before dropping them and they aren't quite dead. They're unreachable only while the `\*` prefix is present; a non-conforming writer that emits a bare `{\bkmkstart …}` falls through to the named set:

```python
src = r'{\rtf1 A{\bkmkstart ch1}B{\bkmkend ch1}C\par}'

# with the entries
'ABC\n'

# with {'bkmkstart', 'bkmkend'} removed from _SKIP_DESTINATIONS
'Ach1Bch1C\n'
```

So the bookmark *names* leak into the body text without them. The `\*`-prefixed form behaves identically either way, as you'd expect:

```python
r'{\rtf1 A{\*\bkmkstart ch1}B{\*\bkmkend ch1}C\par}'  ->  'ABC\n'   # both ways
```

Given the whole point of the fallback is to cope with files that `striprtf` isn't around to handle, I'd lean toward leaving them as cheap insurance against sloppy writers — but it's your call and I'm happy either way. If you'd like them gone I'll send a one-line PR; if they stay, it might be worth a short comment on those two entries saying they exist for the non-`\*` malformed case, so the next reader doesn't reach the same "these look redundant" conclusion and delete them.

Same reasoning applies to a couple of the others in that set (`pict`, `objdata`, `docvar` are conventionally bare, `latentstyles` and `datastore` are conventionally `\*`-prefixed), which is why I kept the named list even though the generic `\*` rule covers a lot of it.

PR #112, opened 2026-08-05
fix(extractor): detect whitespace-separated CJK ToC headers

### LINKED ISSUE #111: [bug] CJK ToC headers with inserted whitespace are not detected
## What happened

`detect_structure()` recognizes the Chinese/Japanese ToC headers `目录`, `目錄`, and `目次` only when their characters are adjacent.

PDF text extraction commonly inserts spacing between CJK glyphs used as display headings. When `pdftotext` emits `目 录` (ASCII space) or `目　录` (U+3000 ideographic space), `has_toc` is `false`. The CLI then warns that no table of contents was detected even though one is present.

## What you expected

Whitespace-separated CJK ToC display headings should be recognized while retaining the whole-line anchoring that prevents inline false positives.

## Source document

- Format: PDF, extracted with `pdftotext`
- Pages / size: 142 pages / 1.5 MB
- Language: Simplified Chinese
- Does `python3 scripts/extract.py --check` show the relevant extractor installed? yes

The original document is copyrighted and is not needed to reproduce the bug.

## Repro

No book file is required:

```python
from book_to_skill.utils import detect_structure

samples = [
    "目录\n第1章 开始\n第2章 进阶",
    "目 录\n第1章 开始\n第2章 进阶",   # ASCII space
    "目　录\n第1章 开始\n第2章 进阶",  # U+3000 ideographic space
]

for text in samples:
    header = text.splitlines()[0]
    print(repr(header), detect_structure(text)["has_toc"])
```

Actual:

```text
'目录' True
'目 录' False
'目　录' False
```

Expected:

```text
'目录' True
'目 录' True
'目　录' True
```

## Root cause

`_TOC_PATTERN` is built by `re.escape()` over literal entries in `_TOC_HEADERS`:

```python
"目录", "目錄", "目次"
```

The generated pattern allows whitespace around the whole header but not between its CJK characters.

## Suggested fix

Allow optional whitespace inside the CJK ToC alternatives, for example:

```regex
目\s*(?:录|錄|次)
```

Alternatively, normalize inter-character whitespace only for the CJK header check. Tests should cover both ASCII space and U+3000 ideographic space so PDF-extractor output remains supported without weakening the inline false-positive guard.

## Environment

- OS: macOS
- Python: 3.9.6
- book-to-skill commit: `c18b0fb2db8b3f3aeb0900ea97db2189764f642c`

aryansk:
## Summary (Required)

Recognize CJK table-of-contents headings even when PDF extraction inserts ASCII, tab, or ideographic whitespace between the characters. The existing whole-line guard remains in place, so inline prose is not reclassified.

## Type of change (Required)

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)

- **Fixes:** PDF text extraction can separate the characters in 目录, 目錄, and 目次, causing otherwise valid table-of-contents headings to be missed.
- **Adds:** Focused coverage for ASCII and ideographic whitespace in simplified Chinese, traditional Chinese, and Japanese forms.

Closes #111

PDF extraction commonly introduces spacing between CJK characters. Those extracted headings should receive the same whole-line treatment as their adjacent-character forms.

The CJK headings use a dedicated regular-expression branch that permits only horizontal ASCII, tab, or ideographic whitespace between the two characters. It remains inside the existing anchored whole-line pattern, preserving the false-positive guard for inline prose.

## Evidence (Required)

- **Tests:** Added a four-case parameterized regression covering `目 录`, `目　录`, `目 次`, and `目　次`.
- **Before / after:** Before this change those four extracted forms were not detected; after the change all four report `has_toc = True`. Existing adjacent-character tests remain green.

```text
274 passed, 1 skipped in 0.47s
ruff check .: All checks passed
validate_skill.py SKILL.md: passed with one existing soft line-count warning
```

There is no UI change. The regression output above covers the extractor behavior directly.

- [x] `ruff check .` is clean

The branch was merged with current `master` at `442aaaa2d21dbe5ae0f7c6a396585ef3a9d2534e`; the PR diff remains limited to the extractor and its regression test.


virgiliojr94:
Thanks for this — I reproduced the bug on current `master` before looking at the patch:

```
detect_structure("目 录\n第一章 开始")["has_toc"]  ->  False
detect_structure("目录\n第一章 开始")["has_toc"]   ->  True
```

So the defect is real: extraction inserts a space (or an ideographic space) between the two characters and the ToC header stops matching. Your `目[ \t　]*(?:录|錄|次)` fix is the right shape — it keeps the end-of-line anchor, so the inline false positive the pattern was designed to avoid still can't match. I rebased it locally on `master`: 278 tests pass, `ruff` clean, and the repro above flips to `True`.

One thing blocks the merge, and it's not the code: **the PR description check is failing** — the body doesn't fill the template (`## Summary`, `## Type of change`, `## Evidence`, and every Checklist item ticked). That gate is a required status check, so I can't merge past it, and I'd rather not rewrite your description for you.

Fill in the template and it goes in as-is. If it helps, the evidence is already written above — feel free to paste it.

This closes #111.

virgiliojr94:
Correction to my previous comment — the description gate is **not** your problem, and I caused the confusion.

I relaxed this repository's fork-PR approval policy so contributors stop waiting on me to press a button for CI to run, and then approved every workflow run that had been sitting queued. Those old runs re-executed carrying the event payload they were created with — which, for this PR, was the body as it existed when the PR was first opened. That's the failing "PR description check" you see. The run from your current description passes:

```
PR description check (PR)   SUCCESS   (current)
PR description check (PR)   FAILURE   (replayed from the original event payload)
```

Your template is filled in correctly. Nothing for you to do — sorry for the noise. Merging.

aryansk:
Thanks for the clarification — good to know the description gate was a replay artifact and the current run passes. Appreciate the merge.

PR #101, opened 2026-08-04
feat(pdf): apply pdftotext cleanup to pypdf and pdfminer paths too

dex0shubham:
## Summary (Required)

PDF cleanup (dehyphenation + repeated header/footer/page-number removal) was only
wired into the `pdftotext` path (#77). The `pypdf` and `pdfminer` extractors still
returned raw text with hyphen-wraps and running headers. This reuses the existing
`clean_pdftotext` on those two paths so PDF text is cleaned consistently no matter
which extractor runs.

## Type of change (Required)
- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ Feature (non-breaking change that adds capability)
- [x] ⚡ Performance (faster / cheaper, no behavior change)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling
- [ ] 💥 Breaking change (changes existing behavior or a public interface)

## What it does (Required)
- **Adds:** `extract_with_pypdf` now joins page text with a form feed and passes it
  through `clean_pdftotext`; `extract_with_pdfminer` (whose output is already
  form-feed delimited) passes its text through `clean_pdftotext`. No new logic —
  reuses the function added in #77.
- **Improves:** consistent PDF cleanup across all three extraction paths; `pypdf`
  is the pip `book-to-skill[pdf]` default, so this is a live path.

## Motivation & context (Required)
#77 cleaned `pdftotext` output only. A book extracted via `pypdf`/`pdfminer` (the
fallbacks, and `pypdf` the pip default) kept hyphen-wrapped words and repeated
headers/footers — the same token/noise problem #77 fixed, just on the other paths.
Closes #n/a (no tracking issue; follow-up to #77).

## How it works (Required for anything non-trivial)
`clean_pdftotext` already handles both cases: with `\f` page boundaries it strips
repeated edge headers/footers and edge page numbers; without them it just
dehyphenates. `pypdf` gives per-page parts, so joining with `\f` lets it get the
full page-aware cleanup; `pdfminer.high_level.extract_text` is already `\f`
delimited. Conservative guards (edges only, ≥3 pages) are unchanged. `docling`
(markdown output) is left untouched.

## Evidence (Required)
- **Tests:** `TestPdfLibsCleanup` (2) mock `pypdf` / `pdfminer` to emit
  hyphen-wrapped, header-repeated pages and assert the returned text is
[...]

PR #113, opened 2026-08-05
fix(pdf): strip running headers only at page edges

Hotragn:
> **Rescoped after #119 merged.** This PR originally also fixed the Roman-numeral false positive (`civil`, `dim`, `mild`…). @Zhengzhongjie fixed that independently in #119 with a cleaner approach, so that commit and the shared `roman.py` module it required are **dropped**. What remains are two failures #119 does not touch, both still reproducing on current `master`. Force-pushed onto `29991b2`.

## Summary

`clean_pdftotext` collects boilerplate candidates from each page's first and last non-blank line, but then removed those strings from **every** line of the page. When a running header repeats the section title — ordinary typesetting — the genuine mid-page heading was deleted along with the headers. Separately, a single-line page voted twice toward the boilerplate threshold.

## Type of change

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does

- **Fixes (1):** boilerplate removal ran over all lines although candidates are only ever *collected* from `nb[0]` / `nb[-1]`. Confined to the page edges.
- **Fixes (2):** on a one-line page `nb[0] is nb[-1]`, so that page cast two votes toward the `> len(pages) / 2` threshold. Counted once now.

## Motivation & context

Closes n/a — found while auditing the #77 cleanup; no existing issue.

Both are silent content loss. Nothing is logged, and `metadata.json` shows no sign a line went missing — so a dropped heading is invisible in the output and downstream (`detect_structure`, the Step 3 chapter map, the generated topic index) simply sees a book with fewer sections.

Since #101 the cleanup also runs on the `pypdf` and `pdfminer` paths, so this now affects every PDF extractor rather than just `pdftotext`.

Failure (1) is the one that hurts most: a running header that repeats the chapter or section title is extremely common, and that is exactly the string chapter detection needs.

## How it works

```python
for i, ln in enumerate(lines):
    if i in (first, last):          # edges only — where boiler was detected
        s = ln.strip()
        if s in boiler or _PDF_PAGE_NUM.match(s):
            continue
    kept.append(ln)
```

The change simply aligns *removal* with *detection*. A running header or footer is by definition at a page edge; a match anywhere else is a coincidence, and treating it as boilerplate is what destroys the heading.

For (2), `edge[nb[-1]] += 1` is guarded by `if len(nb) > 1`.

`_PDF_PAGE_NUM` is untouched — that is #119's, and this PR keeps it working (asserted below).

## Evidence

**Failure 1** — 4 running headers "Reliability" plus 1 genuine mid-page heading of the same string:

```
BEFORE  occurrences in output: 0      <- the real heading is gone too
AFTER   occurrences in output: 1      <- heading kept, all 4 headers still removed
        page numbers 42-45 removed: True
```

**Failure 2** — 4 pages, two of which contain only `PART ONE` (2 votes; threshold is `> 2`):

```
BEFORE  "PART ONE" count: 0    <- both divider pages deleted
AFTER   "PART ONE" count: 2
```

**Tests** — new `tests/test_pdftotext_edge_only_boilerplate.py`, 13 cases in three groups:

- `TestBoilerplateRemovalIsEdgeOnly` — the mid-page heading survives, surrounding body intact, running headers still removed, page numbers still removed, plus a plain running-header and a footer case to prove stripping power is unchanged.
- `TestSingleLinePageVoting` — the part-divider page survives, but a single-line page genuinely over the threshold (3 of 4) is still stripped.
- `TestExistingBehaviourPreserved` — regression net for #77, #101 and **#119**: dehyphenation, short documents, mid-page bare numbers, `CIVIL`/`MIX`/`VIVID` one-word lines still surviving, and Roman front-matter numbers still stripped.

RED against `master` (`29991b2`): **2 failed, 11 passed** — exactly the two bugs, with the other 11 passing because they assert behaviour that already works. GREEN with the fix:

```
$ python -m pytest tests/ -q
360 passed, 3 skipped in 5.88s

$ ruff check .
All checks passed!
```

## Screenshots / output

The before/after counts above are the output; this is about which lines survive, which a text diff shows precisely.

## Checklist

- [x] `ruff check .` is clean

## Notes for reviewers

Credit where due: #119 was the better fix for the numeral half of the original PR, and I've dropped mine rather than argue for it. The `roman.py` module went with it — the remaining changes need no shared parser, so that churn is gone too.

One judgement call worth checking: confining removal to the edges means a running header that also appears mid-page is now kept there. I think that is right — a mid-page occurrence is content, not furniture — but if you would rather keep the old "remove everywhere" behaviour and instead special-case only the *first* occurrence, say so and I will rework it.

Still open from my earlier comment on #71, unrelated to this PR: `validate_docx_xml_safety` decodes every archive member five times, which is a ~10× constant factor on top of that PR's size caps. Happy to send it separately once #71 lands or stalls.

### DIFF
--- a/book_to_skill/parsers/pdf.py
+++ b/book_to_skill/parsers/pdf.py
@@ -39,7 +39,13 @@ def clean_pdftotext(text: str) -> str:
             nb = [ln.strip() for ln in p.splitlines() if ln.strip()]
             if nb:
                 edge[nb[0]] += 1
-                edge[nb[-1]] += 1
+                # On a single-line page the first and last line are the same
+                # line. Counting it twice would let one page cast two votes
+                # toward the "more than half the pages" threshold below, so a
+                # part-divider page occurring twice in four pages would reach 4
+                # votes instead of 2 and be stripped as boilerplate.
+                if len(nb) > 1:
+                    edge[nb[-1]] += 1
         boiler = {ln for ln, c in edge.items() if c > len(pages) / 2}
         kept = []
         for p in pages:
@@ -48,12 +54,16 @@ def clean_pdftotext(text: str) -> str:
             first = nb_idx[0] if nb_idx else None
             last = nb_idx[-1] if nb_idx else None
             for i, ln in enumerate(lines):
-                s = ln.strip()
-                if s in boiler:
-                    continue
-                # Drop a bare page number only at a page edge (varies per page).
-                if i in (first, last) and _PDF_PAGE_NUM.match(s):
-                    continue
+                # Running headers/footers and page numbers only ever occur at a
+                # page edge -- which is also the only place `boiler` is
+                # collected from. Removing a boilerplate string from every line
+                # meant that when a running header repeated the section title
+                # (common typesetting), the genuine mid-page heading was deleted
+                # along with the headers.
+                if i in (first, last):
+                    s = ln.strip()
+                    if s in boiler or _PDF_PAGE_NUM.match(s):
+                        continue
                 kept.append(ln)
         text = "\n".join(kept)
     else:
--- a/tests/test_pdftotext_edge_only_boilerplate.py
+++ b/tests/test_pdftotext_edge_only_boilerplate.py
@@ -1,0 +1,159 @@
+"""Running-header removal must be confined to page edges.
+
+`clean_pdftotext` collects boilerplate candidates from the first and last
+non-blank line of each page (`nb[0]` / `nb[-1]`), but then removed every
+occurrence of those strings from *every* line of the page. Books very commonly
+set the running header to the chapter or section title, so the header string and
+a genuine heading are byte-identical — and the genuine heading was deleted along
+with the headers.
+
+Separately, a page whose only content is one line counted that line twice toward
+the "repeated on more than half the pages" threshold, because it is both `nb[0]`
+and `nb[-1]`.
+
+Both failures delete real text, and they are silent: nothing is logged, and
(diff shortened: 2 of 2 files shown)


virgiliojr94:
Thank you for this — the core finding is correct, and I confirmed it on `master` before reviewing:

```
clean_pdftotext(pages with "MIX" / "CIVIL" / "VIVID" alone on a page edge)
  ->  all three silently deleted
```

Silent content loss is the worst failure mode this project has, so that one needed to land fast. It has: #119 fixed the same root cause with a canonical-numeral pattern and shipped in 6c07a96. I went with that one because it was the smaller change for that single defect — not because this PR was worse.

What I don't want to lose is the **other two findings in here**, which #119 does not cover and which I believe are also real:

1. **The double vote on a single-line page.** `edge[nb[0]] += 1` and `edge[nb[-1]] += 1` are the same line when a page has one non-blank line, so one page contributes two votes toward the `> len(pages)/2` boilerplate threshold.
2. **Boilerplate removal not restricted to page edges.** `boiler` is only ever *collected* from edges, but it was *applied* to every line — so a running header that repeats the chapter title also deleted that title where it legitimately appears in the body.

Both are one-defect-sized, both are testable in isolation, and (2) is another instance of the same silent-deletion class. Would you open them as two separate PRs against current `master`? They'll get reviewed quickly.

The `roman.py` extraction I'd leave out unless it's needed on its own merits — `utils.py` already has `_roman_to_int`, so if the dedup is worth doing it deserves its own PR and its own argument.

PR #117, opened 2026-08-06
fix: force UTF-8 encoding for pdftotext and metadata output on Windows

carlosarisco55-max:
## Summary (Required)
`pdftotext` (poppler) writes non-ASCII output using the platform's legacy 8-bit encoding by default on Windows, not UTF-8. Since `extract_with_pdftotext` decodes stdout as UTF-8, this produced mojibake for any accented/non-ASCII text extracted from PDFs. This PR passes `-enc UTF-8` explicitly to `pdftotext` so its output encoding matches what we decode it as.

## Type of change (Required)

## What it does (Required)
- **Fixes:** garbled (mojibake) output from `extract_with_pdftotext` on Windows for any non-ASCII text (á/é/í/ó/ú/ñ/¿/¡ etc.), caused by `pdftotext` defaulting to a legacy 8-bit encoding while the caller decodes stdout as UTF-8.

n/a — found while running extraction on a Spanish-language PDF locally on Windows; no existing issue tracked it.

One-line fix: add `-enc UTF-8` to the `pdftotext` subprocess invocation in `extract_with_pdftotext` so its output encoding matches the `encoding="utf-8"` already used to decode `stdout`. No behavior change on platforms where `pdftotext`'s default was already UTF-8.

(This PR originally also touched `metadata.json` encoding — dropped after rebasing, since that was already fixed upstream in #105.)

## Evidence (Required)
- **Tests:** extended `TestPdftotextEncoding::test_pdftotext_decodes_as_utf8` in `tests/test_book_to_skill.py` to assert `-enc UTF-8` is actually present in the `pdftotext` command, not just that stdout decoding is UTF-8 (the previous test didn't cover the fix itself). `pytest -q` → 270 passed, 1 skipped.
- **Before / after:** reproduced and verified locally on Windows, extraction on a Spanish-language PDF before/after the fix:

```
Before: M�dulo 2: Estrategia y t�cticas del marketing digital
After:  Módulo 2: Estrategia y tácticas del marketing digital
```

Also observed table-of-contents detection improve as a side effect — `has_toc` flipped from `false` to `true` on the same file, since the ToC heading regex depends on correctly-decoded accented headings like "ÍNDICE".

- [x] `ruff check .` is clean

`SKILL.md` is untouched by this PR, so `tools/validate_skill.py` doesn't apply here. Happy to adjust the approach (e.g. detecting encoding instead of forcing it) if you'd prefer.

### DIFF
--- a/book_to_skill/parsers/pdf.py
+++ b/book_to_skill/parsers/pdf.py
@@ -53,7 +53,7 @@ def extract_with_pdftotext(pdf_path: str) -> str | None:
     try:
         pdf_path = os.path.abspath(pdf_path)
         result = subprocess.run(
-            ["pdftotext", "-layout", pdf_path, "-"],
+            ["pdftotext", "-layout", "-enc", "UTF-8", pdf_path, "-"],
             capture_output=True, text=True, timeout=120,
             encoding="utf-8", errors="replace",
         )
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -736,7 +736,7 @@ def main():
         **consolidated_structure,
     }
     
-    OUTPUT_META.write_text(json.dumps(metadata, indent=2, ensure_ascii=False))
+    OUTPUT_META.write_text(json.dumps(metadata, indent=2, ensure_ascii=False), encoding="utf-8")
     
     page_line = f"   Total Pages: {total_pages}"
     print("\nExtraction complete:")
--- a/tests/test_book_to_skill.py
+++ b/tests/test_book_to_skill.py
@@ -1496,6 +1496,7 @@ class TestPdftotextEncoding:
         monkeypatch.setattr(pdf_parser.shutil, "which", lambda name: "/usr/bin/pdftotext")
 
         def fake_run(cmd, **kwargs):
+            captured["cmd"] = cmd
             captured.update(kwargs)
             return _Result()
 
@@ -1504,6 +1505,12 @@ class TestPdftotextEncoding:
         assert pdf_parser.extract_with_pdftotext("x.pdf") == "Café — naïve"
         assert captured.get("encoding") == "utf-8"
         assert captured.get("errors") == "replace"
+        # pdftotext itself must be told to emit UTF-8, not the platform's
+        # legacy 8-bit default (this is what actually avoids mojibake on
+        # Windows; decoding stdout as UTF-8 alone isn't enough).
+        cmd = captured.get("cmd") or []
+        enc_idx = cmd.index("-enc")
+        assert cmd[enc_idx + 1] == "UTF-8"
 
 
 class TestPdftotextCleanup:


carlosarisco55-max:
Quick update: rebased on latest \`master\` — the \`metadata.json\` encoding fix from this PR was already covered by #105, so I dropped that part and kept only the \`pdftotext -enc UTF-8\` fix, which is still needed (confirmed it's not addressed elsewhere on master). Should be conflict-free now.

virgiliojr94:
Thanks for this. Before merging I tried to reproduce the failure it describes, and I couldn't — so I want to put the measurement in front of you rather than quietly sit on the PR.

The premise in both this PR and its test comment is that `pdftotext` emits Latin-1 by default and therefore drops non-Latin-1 characters. On poppler 24.02 that isn't what the tool does:

```
$ man pdftotext
   -enc encoding-name
          Sets the encoding to use for text output. This defaults to "UTF-8".

$ pdftotext -layout enc.pdf -            # no -enc
em dash — quote “x” ellipsis … euro €

$ pdftotext -layout -enc UTF-8 enc.pdf - # with -enc
em dash — quote “x” ellipsis … euro €
```

Em dash, curly quotes, ellipsis and euro are all outside Latin-1, and all survive without the flag. So on a current poppler this patch is a no-op, and the stated bug isn't the bug.

That doesn't automatically make it worthless. What I could **not** establish is which poppler release changed that default, or whether the Windows builds people actually install (xpdf-derived, or older poppler bundles) still default to something 8-bit. If they do, `-enc UTF-8` is two arguments that make the guarantee explicit instead of inherited — cheap insurance, and I'd take it.

What I'd need to merge it is the justification corrected to match reality: not "pdftotext defaults to Latin-1", but "we pin the output encoding rather than depend on the build's default", ideally with the Windows version you hit this on and what its output actually looked like. If it turns out your poppler really does emit mojibake, that's a much stronger PR than the current one and I'd like to see that evidence.

Note #124 proposes the identical one-line change with a cleaner dedicated test. Yours came first, so I'm keeping this one open and closing that one — if you update this PR, please lift `test_pdftotext_requests_utf8_output` from #124 and credit @kang3979 as co-author.

virgiliojr94:
Unrelated to the review above: the red `PR description check` on this PR is my fault, not yours. I relaxed the fork-PR approval policy and then approved every queued workflow run — the old ones replayed with the event payload they were created with, i.e. your description as it was when the PR was opened. The run against your current body passes. Nothing to fix there.

CI will now run on your pushes without waiting for my approval.

virgiliojr94:
Issue #192 provided the exact reproduction: on Windows with xpdf 4.06 on PATH, `pdftotext` defaults to Latin-1, causing mojibake on accented text and breaking ToC detection. Passing `-enc UTF-8` fixes it completely. Please rebase your branch on `master` and ensure `-enc UTF-8` is passed in both `extract_with_pdftotext` and `looks_image_only` in `book_to_skill/parsers/pdf.py`. Ready to merge once updated.

virgiliojr94:
Landed on master with original commit author attribution. Thanks @carlosarisco55-max and @Pm9158!

PR #124, opened 2026-08-08
fix(extractor): request UTF-8 output from pdftotext

kang3979:
## Problem

`pdftotext` writes **Latin-1** by default. `extract_with_pdftotext()` decodes the subprocess as UTF-8, but by then the damage is done — every non-Latin-1 codepoint was already dropped by `pdftotext` itself. Hangul, Kana, Han and Cyrillic silently disappear.

The failure is quiet rather than loud: the run succeeds, `chapters_detected` looks sane, and the word count is merely *small* rather than zero. `pdftotext` is tried first and its output is accepted because it is non-empty, so the healthier `pypdf` fallback never runs.

## Fix

Pass `-enc UTF-8`.

```diff
-["pdftotext", "-layout", pdf_path, "-"],
+["pdftotext", "-layout", "-enc", "UTF-8", pdf_path, "-"],
```

## Measurement

A 2,530-page corpus of seven Korean qualitative-methods books, `--mode text`:

| | words | words/page |
|---|---|---|
| before | 141,787 | ~56 |
| after | **539,750** | ~213 |

3.8×. These are not scanned-only PDFs — `pypdf` extracts the same pages fine, which is exactly why the loss is easy to miss.

Sample of the same three pages, before and after:

```
before:  ' ...  '`.I  r  `.`  ,`  Hb  '  140
after:   140 질적 연구 이해 논문 초보자를 위한 가이드
         과정을 거쳐 진행되었는지를 들어야 할 때가 있다. …
```

## Test

`TestPdftotextEncoding` already existed but asserted only the Python-side decode kwargs (`encoding`, `errors`) — never the argv — so it passed both before and after. Added `test_pdftotext_requests_utf8_output`, which captures the argv and asserts `-enc UTF-8` is present.

Verified it actually discriminates:

```
# with the old argv
E  AssertionError: pdftotext invoked without -enc:
   ['pdftotext', '-layout', 'x.pdf', '-']
1 failed

# with the fix
8 passed
```

Full suite: `272 passed`, `ruff check .` clean.

## Notes

- Affects every non-Latin-1 script, not just Korean; Korean is simply where I hit it.
- Independent of #111 (CJK ToC headers with inserted whitespace) — that one is about heading detection *after* text is extracted, this is about the text never arriving.

### DIFF
--- a/book_to_skill/parsers/pdf.py
+++ b/book_to_skill/parsers/pdf.py
@@ -53,7 +53,7 @@ def extract_with_pdftotext(pdf_path: str) -> str | None:
     try:
         pdf_path = os.path.abspath(pdf_path)
         result = subprocess.run(
-            ["pdftotext", "-layout", pdf_path, "-"],
+            ["pdftotext", "-layout", "-enc", "UTF-8", pdf_path, "-"],
             capture_output=True, text=True, timeout=120,
             encoding="utf-8", errors="replace",
         )
--- a/tests/test_book_to_skill.py
+++ b/tests/test_book_to_skill.py
@@ -1484,7 +1484,7 @@ class TestTextEncodingDetection:
 
 
 class TestPdftotextEncoding:
-    """pdftotext output (UTF-8) is decoded as UTF-8, not the locale encoding."""
+    """pdftotext is asked for UTF-8 output, and that output is decoded as UTF-8."""
 
     def test_pdftotext_decodes_as_utf8(self, monkeypatch):
         captured = {}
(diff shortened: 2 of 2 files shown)


virgiliojr94:
Thanks for this, and sorry to close it — the change is identical to #117, which was opened two days earlier, so I'm consolidating on that one for precedence rather than on merit.

Your test is the better of the two: `test_pdftotext_requests_utf8_output` asserts the argv contract directly instead of bolting an assertion onto the decoding test. I've asked @carlosarisco55-max to lift it into #117 with you credited as co-author, so it isn't lost.

One thing worth flagging on both PRs: I tried to reproduce the failure and couldn't. On poppler 24.02, `man pdftotext` documents `-enc` as defaulting to UTF-8, and em dash / curly quotes / ellipsis / euro — all outside Latin-1 — round-trip correctly without the flag. Full measurement is in #117. The change may still be worth taking as an explicit pin rather than an inherited default, but not on the "defaults to Latin-1" reasoning.

If you hit real mojibake on a specific platform, please do open an issue with the poppler/xpdf version and the mangled output — that's the evidence that would settle it.

PR #125, opened 2026-08-08
feat(skill): default generated skills to ~/.agents/skills with Claude Code symlink

n0an:
## Summary (Required)
Split out of #100 per review. Personal (user-level) installs of **generated** book skills default to the cross-agent root `~/.agents/skills/<slug>/`, so one generated copy serves Copilot CLI, Amp, and Codex natively. Step 10 attempts a symlink at `~/.claude/skills/<slug>` when running under Claude Code (the one host that does not scan the cross-agent root) and **reads the link back before reporting anything about it**. Hermes Agent keeps its own personal root, since it partitions personal skills by category and does not scan the cross-agent root either.

## Type of change (Required)
- [x] ✨ Feature (non-breaking change that adds capability)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Adds:** `SKILL.md` Step 5 collapses the per-host personal-root tables into one default (`~/.agents/skills`), keeping host-private and project-local roots available on explicit request, and keeping the Hermes row as its own personal root. Step 10 attempts the Claude Code symlink, then resolves it into `CLAUDE_STATUS` (`linked` / `copy` / `absent` / `skipped-realdir`) and fills the report's "Discoverable by" line from that read-back, never from the fact that `ln` ran. A pre-existing real directory at `~/.claude/skills/<slug>` triggers a migration offer instead of a silent replace, and is guarded before `ln -sfn` so the link cannot nest inside the stale copy. The report gains a relocation block whose `mv` targets the final skill directory (`<dest_root>/<skill_name>`), so the printed command never breaks the symlink the run just created, plus one line answering the out-of-cwd permission-prompt case. README "What it generates" and the `docs/how-it-works.md` pipeline diagram match, including the honest wording that the symlink is attempted and verified rather than asserted.

Closes n/a. A book converted under Claude Code lands in `~/.claude/skills/` and is invisible to Copilot CLI and Amp on the same machine (and vice versa) - the converter's own value proposition is "one skill, any host", but its output today is host-locked. The cross-agent root plus one verified symlink fixes that with zero divergence risk (one physical copy).

**On the "ask once, don't silently default" rule** (the original review concern): this PR narrows the ask, it does not remove user control. That rule exists because the per-host roots were *alternatives* - picking one silently meant guessing wrong for the user's other agents. For the four hosts that reach `~/.agents/skills` natively or through a verified link, there is no longer a wrong guess to make for the personal-install case, so the question becomes ceremony. The ask remains wherever a real choice exists: project-local output, host-private roots, and the migration of a pre-existing real directory are all explicit-request or explicit-confirm paths. Step 5 also keeps one carve-out so the default does not invent a convention in someone else's house: if `~/.agents/skills` does not exist and the host's private root already holds skills, the private root wins and the report says why.

**Hermes Agent is the exception, deliberately.** Its personal root is `$HERMES_HOME/skills/<category>`, partitioned by category, and it does not scan the cross-agent root, so "one root every host can reach" is not true of it. Rather than send Hermes users' skills to a root Hermes never reads, a personal Hermes install uses the Hermes row and takes no symlink, and the report names Hermes alone for that destination with no cross-agent claim. Extending the verified-link treatment to Hermes (with `hermes skills list` as the read-back) is a reasonable follow-up, but it is unverified from here and out of scope for this PR unless you want it in.

Instruction-level change to `SKILL.md` only - no Python code paths. The symlink is attempted after generation and the Step 9.5 scan (never mid-write), and only when the host is Claude Code and the destination is the default personal root. It is then read back (`test -L` plus a `readlink` compare) and the report states what is on disk, because on Windows/MSYS `ln -s` may copy instead of link, may need Developer Mode, or may be absent entirely - and a report that claims discovery it cannot see is how one copy silently rots on the next Update/Fold-in. A missing or copied link does not hard-fail the run: the skill exists at the hub and every other host still finds it. Rejected alternative: copying the skill into both roots - two copies drift after the first Update/Fold-in run.

## Evidence (Required)
- **Tests:** `pytest tests/ -q` - 622 passed, 5 skipped on a clean checkout; `ruff check --select E9,F --target-version py310 book_to_skill/ scripts/ tests/ tools/` clean; `python3 tools/validate_skill.py SKILL.md` exit 0 (pre-existing body-length soft warning only).
- **Before / after:** before - a skill generated under Claude Code is invisible to Copilot CLI/Amp. After - verified live with a freshly generated 35-chapter skill:

```
$ ls -l ~/.claude/skills/clean-architecture-book
lrwxr-xr-x ... ~/.claude/skills/clean-architecture-book -> /Users/anton/.agents/skills/clean-architecture-book

$ ls ~/.agents/skills/clean-architecture-book
SKILL.md  chapters/  cheatsheet.md  glossary.md  patterns.md
# Claude Code loads it through the symlink; Copilot CLI and Amp read the physical copy natively
```

- **Read-back, all five paths:** the Step 10 block was extracted verbatim and driven through `ln` working, `ln` missing, `ln` copying instead of linking (the MSYS default), a pre-existing real directory, and a stale symlink pointing elsewhere. It reports `linked` / `copy` / `absent` / `skipped-realdir` correctly in each, always from the read-back. Real Windows and the `mklink /J` junction remain unverified and are labeled as such in the file.
- **Relocation line, both readings:** run against a scratch `$HOME`. With the move target read as a root the printed command breaks the link it just created (`cat: ~/.claude/skills/mybook/SKILL.md: No such file or directory`); with the target as the final skill directory the link resolves to `~/.copilot/skills/mybook` and the file reads back. Hence `<dest_root>/<skill_name>` in the template.

- [x] `ruff check .` is clean

- Split from #100 as requested; #100 is docs-only again. No `CHANGELOG.md` edit (git-cliff per #104).
- #102 has since merged; this branch was rebased over it, and the Step 10/11 regions no longer conflict.
- Rebased again over #202 (Hermes Agent), which landed in the same Step 5 table this PR collapses. The Hermes handling above is the resolution; the unknown-host prompt keeps `master`'s wording so `test_hermes_is_in_unknown_host_prompt_and_reload_guidance` passes unmodified.
- No new pytest cases for the default itself: the change is instruction-level (`SKILL.md`); no extractor or tooling code is affected. The Hermes contract tests in `tests/test_hermes_host_support.py` cover the row and prompt that this PR touches.


n0an:
@virgiliojr94 heads-up: addressed all five automated review points (Codex + Copilot) in `7028dc8`:

- **Symlink nesting bug (Codex P2):** Step 10 now guards the real-directory case - if `~/.claude/skills/<skill_name>` remains a real directory (migration declined), the symlink is skipped with an explicit report line instead of `ln -sfn` nesting the link inside it and leaving Claude Code on the stale copy.
- **Discovery claims (Codex P2):** the report's "Discoverable by" line is now filled conditionally from the destination actually used and whether the symlink was created - no more claiming Copilot/Amp discovery for a `~/.claude/skills`-only install.
- **Placeholder ambiguity (Copilot):** `<slug>` → `<skill_name>` for the skill folder in SKILL.md.
- **Overpromising docs (Copilot ×2):** README and the `docs/how-it-works.md` diagram now say the symlink is added when running under Claude Code, not unconditionally.

This PR is ready for your call on the default-root question whenever you get to it.

virgiliojr94:
Decision on the default root: **I'm taking it.** Below is the reasoning, two changes I want first, and one question I'd like your read on before merge.

## Why I'm reversing my own rule

"When more than one valid root exists, ask the user once and remember — do not silently default" is my line, and I wrote it to protect user agency. Working through this PR I think it protects the wrong thing at the wrong moment. The question lands in the middle of an operation that costs 155K–391K tokens and about a dollar, and in nearly every case the answer is the same. Agency that costs a prompt on every run, to confirm the obvious, is friction wearing agency's clothes.

What changed my mind on the specific destination is that the ecosystem already settled it. I ran the official `skills` CLI on a clean machine today:

```
$ npx skills add <repo>
✓ ~/.agents/skills/<slug>
  universal: Codex, Gemini CLI, GitHub Copilot, OpenCode, Amp +12 more
  symlinked: Claude Code
```

One copy at the cross-agent root, symlink for Claude Code — the exact shape of this PR. Users who install skills any other way already have this layout. And the two layouts coexist fine in practice: on my machine `~/.claude/skills` now holds 15 real directories plus one symlink into `~/.agents/skills`, with nothing confused about it.

Two things in here are already right and I want to say so before the asks: the conditional "Discoverable by" line — including the case where it must print "**NOT** Claude Code — migrate first" — is the honest-reporting posture this project runs on. And the real-directory guard before `ln -sfn` is the correct fix for a bug that would otherwise nest the link inside the stale copy and leave Claude Code loading it.

## Two changes before merge

**1. Give the default an undo, and print it.** This is the one that makes the whole thing safe. A default is not dangerous because it was chosen silently; it is dangerous when the cost of it being wrong is high. So make it cheap: alongside the "Discoverable by" line, print the move command for the destination actually used, e.g.

```
Move it:  mv ~/.agents/skills/<skill_name> <dest> && ln -sfn <dest> ~/.claude/skills/<skill_name>
```

with the three cases spelled out (hub + symlink / host-private / project-local), so the printed command is correct for the path taken and never breaks the symlink it just created. That single line is what replaces the question I'm removing — the user still decides, they just decide after seeing the result instead of before, and at no cost to the run.

**2. One sentence of fingerprinting, not the table back.** Defaulting to the hub is right when the hub is where skills live. It is presumptuous when the user has never had one: creating `~/.agents/skills` on a machine whose only populated root is `~/.copilot/skills` invents a convention in someone else's house. Rule, one sentence:

> If `~/.agents/skills` does not exist **and** the host's private root already contains skills, use the private root and say why in the report.

Deliberately minimal. I do not want the eight-root probing table coming back through the side door — that weight is exactly what this PR is removing.

## The question I'd like your read on

Adopting the CLI's default means the converter is no longer the thing deciding where skills live; it is a consumer of a convention someone else maintains. If that is true, why do we carry a table of eight roots at all?

When the `skills` CLI is present, we could hand the generated folder to it and let it place and symlink — deleting most of Step 5 rather than restating it. That is a subtraction, which is the direction I want this file moving in anyway (it is 690 lines against a 500-line guideline, and that is my debt, not yours).

I'm not blocking on it and I don't want it in this PR. But you have thought about cross-agent installs more than anyone here, so: does delegating placement to the CLI when it exists strike you as the right next step, or does the fallback path end up costing more than the table it replaces?

Make the two changes above and I'll merge.

virgiliojr94:
New information that changes what I asked you for above, so flagging it before you spend time on the two changes.

#137 came back with the detail I was missing. A Windows user on Claude Code 2.1.220, working in `~/Desktop/weekly`, ran a conversion. `~/.claude/skills` exists on that machine, so Step 5's rule 1 picked it without asking — and because that path is **outside the working directory**, the host asked for authorization on every file the generator wrote. The permission prompts were not a misconfiguration; they are what the default produces on a host that gates writes outside the cwd.

That is a problem this PR does not solve. `~/.agents/skills` is also outside the working directory, and the Claude Code symlink means a second write outside it — so on that user's machine the new default trades one out-of-cwd destination for two.

I still think the cross-agent root is the right default for the reason we discussed: one copy, every host, and the `skills` CLI already established the layout. The case I now cannot answer is what should happen when the host will clearly gate every write — do we accept a prompt per file, or does the run notice and offer the project-local root once?

I do not want that question answered with a new probe or another table; that is the weight this PR is removing. Something closer to: the destination is announced in the Step 10 report either way, and if the user says the writes are being blocked or prompted, the answer is one line telling them to ask for the project-local root and re-run.

Two things before you touch anything:

1. **Hold the two changes I asked for** (the `Move it:` line and the one-sentence fingerprint) until this is settled — the `Move it:` line may end up being the same mechanism that answers this, in which case it should be worded once, for both cases.
2. **Tell me what you think.** You have thought about cross-agent installs more than anyone here, and you may well have hit the prompt-per-file behaviour yourself. If the answer is "accept the prompts, the default is still right", say so and I will take it as is and answer #137 separately — I am not looking to expand this PR, only to avoid shipping a default that makes one class of user worse off without us having said so out loud.

n0an:
Holding the two changes as asked. Here is my read, and one gap in this PR that #137 exposes and I had missed.

## The arithmetic is smaller than it looks

Two destinations, yes. But what that user pays is one authorization prompt per write, and that count goes from N to N+1, not from N to 2N. The chapter files land in `~/.agents/skills/<slug>/` exactly as they land in `~/.claude/skills/<slug>/` today; the symlink is one additional operation at the end of the run. For a 35-chapter book that is roughly 40 prompts either way, plus one.

So this PR does not make his case materially worse. It fails to fix it, which is a different and much weaker thing, and I do not think it should be what blocks the default.

## My answer: accept the prompts, the default is still right

Take #125 as is. The prompt-per-file friction is a host policy about the working directory, not a property of which out-of-cwd path we choose. Every personal-scope install is out-of-cwd by definition. The only destination that avoids the prompts is project-local, and you have already given the correct reason not to default there: a skill written into whatever folder someone happened to be standing in is discoverable only from that folder, and most conversions are meant to outlive the directory.

Put the other way: if out-of-cwd prompts were disqualifying, the conclusion would not be "prefer `~/.claude/skills` over `~/.agents/skills`" - it would be "no personal-scope default at all", which is a much bigger change than this PR and one neither of us wants.

## The `Move it:` line should answer both cases, which is why you were right to hold it

This is the part you were reaching for, and I agree it wants wording once rather than twice. The destination is announced either way; if the writes are being gated, the answer sits next to the destination that caused it:

```
Installed to:  ~/.agents/skills/<skill_name>
               + symlink at ~/.claude/skills/<skill_name>

Somewhere else?  mv ~/.agents/skills/<skill_name> <dest> \
                   && ln -sfn <dest> ~/.claude/skills/<skill_name>

Prompted for permission on every file? That is your host gating writes outside the
working directory. Say "save it in this project" and re-run to write inside it.
```

One block. Three variants of the `mv` for the three destinations (hub + symlink, host-private, project-local), each correct for the path actually taken and none of them breaking the symlink the run just created. No probe, no table, and the blocked-writes case costs two lines rather than a mechanism.

The one-sentence fingerprint stands unchanged from your version: if `~/.agents/skills` does not exist and the host's private root already contains skills, use the private root and say why in the report.

## The gap: the symlink is the wrong shape on Windows

The #137 reporter is on Windows, and I do not think Step 10's symlink survives there.

`ln` is not a command in PowerShell or cmd. In Git Bash it exists, but the MSYS default for `ln -s` is to **copy** rather than link unless `MSYS=winsymlinks:nativestrict` is set, and a native symlink additionally requires Developer Mode or an elevated shell. So on that platform the step plausibly either fails outright or silently produces a second physical copy - which is the exact divergence the single-copy design exists to prevent. The next Update/Fold-in run would then update one copy and leave the other stale, and the report would still claim Claude Code discovery.

I have not tested this on Windows, so treat it as something to confirm rather than something I have verified. If it holds, the fix is small and matches the posture already in this PR: attempt the link, verify it actually resolved (`test -L`), and on failure report the hub path with an honest line - "your host could not create the symlink; Claude Code will not see this skill until you enable Developer Mode or copy it manually" - rather than printing discovery that is not there. That is the same conditional reporting the "Discoverable by" line already does, extended to one more failure mode.

Say the word and I will push the `Move it:` block, the fingerprint sentence, and the symlink verification together.

## On delegating placement to the `skills` CLI

Yes eventually, no as a hard dependency, and not in the same PR that removes the question.

`npx skills add` wants a repo. Handing it a locally generated folder means either publishing first - which is #102's territory, and installing your own skill should not require pushing it to GitHub - or leaning on whatever local-path support the CLI has, which I would want pinned to a version before the converter's core output path depends on it. The shape worth having is: if the CLI is present and supports a local install, hand off and delete the table; otherwise the current path.

The catch is that the fallback *is* the table. So the subtraction only actually pays off at the point you are willing to require the CLI, and until then you are maintaining both. That is a real simplification but it is a later one, and it should not ride along with a change whose whole argument is that it removes weight.

virgiliojr94:
Send the three. And you found the thing that decides this PR, which I had not seen: the symlink is a claim, not a fact.

**On the arithmetic — you are right and I was sloppy.** N+1 writes, not 2N, and the prompts are the host gating writes outside the cwd, which is a property of where the user asked the skill to go, not of this PR. That disposes of my objection as stated.

**On Windows, the reasoning stands even unverified.** Whether MSYS copies, `ln` is absent, or a native symlink needs Developer Mode, all three failure modes end the same way: `~/.claude/skills/<slug>` exists, the run says "Discoverable by: Claude Code", and the user believes it. The next Update/Fold-in then edits one copy while the other rots — and the report that told them everything was fine is the reason they will not think to check.

So the fix I want is not a better link mechanism, it is the refusal to assert:

1. Create the link.
2. **Read it back** (`test -L`, or resolve the path and compare) before printing anything about it.
3. Fill the "Discoverable by" line from what the read-back found — link, copy, or nothing — never from the fact that the command was issued.

That is platform-agnostic, needs no Windows machine to be correct, and matches what the line already does for the migration-declined case. Please do not hard-fail the run when the link is missing: the skill exists at the hub and other hosts can find it, so the honest report is "written to `~/.agents/skills/<slug>`; Claude Code will not see it until X", not an abort.

**One lead worth a look, marked as unverified by me:** Windows junction points (`mklink /J`) are directory links that do *not* require Developer Mode or elevation, unlike `mklink /D`. If a junction works where a symlink does not, that is the Windows branch of step 1 — but it does not change steps 2 and 3, which are what make the report trustworthy either way. I have no Windows machine and have not tested it; treat it as a lead, not an instruction.

**On delegating to the `skills` CLI:** agreed, and your reason is better than my question. The fallback is the table, so the subtraction only pays at the point of requiring the CLI, and requiring it to install your own locally generated skill is backwards. Dropping it.

Push the `Move it:` block, the fingerprint sentence, and the link verification, and this goes in.

n0an:
Pushed `038bc75` — rebased onto `master` and the three changes are in.

**Rebase.** #102 (Step 11 publishing) landed in the same Step 10/11 region, so the branch was conflicting. Resolved against the current file, and carried #154's **OpenAI Codex** row forward — the collapse now sits on top of the four-host table, Codex in the natively-discovered column. README + the `docs/how-it-works.md` diagram name Codex too, for consistency.

**1. Link read-back (the one that decides it).** Step 10 no longer asserts the symlink. It attempts `ln -sfn`, then reads the link back (`test -L` + `readlink` compare) into `CLAUDE_STATUS` = `linked` / `copy` / `absent` / `skipped-realdir`, and the "Discoverable by" line is filled from that — never from the fact that the command ran. The `copy`/`absent` case (Windows/MSYS copies, or no `ln`) reports honestly that Claude Code will not see it and that a copy drifts on the next Update/Fold-in. No hard-fail: the skill exists at the hub regardless. The `mklink /J` junction is in as an explicitly unverified lead, and it does not change the read-back-then-report rule.

**2. `Somewhere else?` / `Prompted…` block.** In the report template, worded once, with the three `mv` variants (hub+symlink / host-private / project-local), each correct for the path taken so it never breaks the symlink the run just created.

**3. Fingerprint sentence.** One line in Step 5 rule 1: private root wins only if `~/.agents/skills` does not exist **and** the private root already holds skills. No table.

Dropped from earlier discussion, as agreed: no `skills` CLI delegation, no hard-fail on a missing link.

**Gates:** `ruff` clean, `validate_skill.py SKILL.md` exit 0 (the pre-existing 755-line soft warning only), `pytest -q` 476 passed / 1 skipped locally; CI green.

virgiliojr94:
This is nearly there. The read-back change is the right shape: Step 10 now treats the symlink as a fact to verify, guards the existing real-directory case, and does not hard-fail when Claude Code discovery cannot be created. That answers the Windows/MSYS failure mode I was worried about.

One mismatch remains: the top-level docs still state the symlink as a fact:

- README: "also drops a symlink at `~/.claude/skills/<slug>/`"
- `docs/how-it-works.md`: "symlink → above, added when running under Claude Code"

But Step 10 now correctly says the result can be `linked`, `copy`, `absent`, or `skipped-realdir`, and the report must be filled from read-back. The public summary should match that: "attempts a verified symlink" / "Claude Code sees it when the link is verified; otherwise the report says it is not discoverable." Otherwise Windows users are still told the thing that may not be true.

Same class, smaller: the relocation example should not leave `<dest>` ambiguous. If `<dest>` is a root, the relink target is wrong after `mv`; if it is the final skill directory, say that. The generated line can use the actual final path and avoid the ambiguity.

With those wording fixes, I would merge this. #137 should stay open unless you want to close it as "documented workaround, default remains personal"; #125 does not remove the out-of-cwd prompts, it only makes the default cross-agent and honest about Claude discovery.

n0an:
Done in 6c6f1c2 — README and the how-it-works diagram now match Step 10's read-back rule: the symlink is attempted and verified, and the report states the actual result. No claim survives that isn't read back from disk.

virgiliojr94:
The first wording fix is in and correct — README and the how-it-works diagram now say the symlink is attempted and verified, and defer to the report for the result. Nothing in the docs claims discovery that read-back has not confirmed.

The second one did not make it: `6c6f1c2` touches `README.md` and `docs/how-it-works.md` only, so `<dest>` is still undefined, 5× in `SKILL.md` (lines 614–615 in the report template, 637–639 in the fill rules).

It matters because both readings of `mv <src> <dest>` are grammatical and only one works. Running the printed line with `<dest>` as a root:

```
$ mv ~/.agents/skills/mybook ~/.copilot/skills && ln -sfn ~/.copilot/skills ~/.claude/skills/mybook
$ cat ~/.claude/skills/mybook/SKILL.md
cat: .../.claude/skills/mybook/SKILL.md: No such file or directory
```

The link resolves to the root, the skill sits at `…/skills/mybook/mybook`, and the relocation line has broken the symlink the run just created — the one thing the fill rule above it says it must never do. With `<dest>` as the final skill directory (`~/.copilot/skills/mybook`) the same two commands work and the skill reads back fine.

Either name it as the final skill directory in the template, or have the generated line print the actual final path so there is nothing to interpret.

Everything else checks out on my side: no new URLs or dependencies, no CI/workflow surface, no invisible or bidi characters in the added text, `CHANGELOG.md` untouched, and it still merges cleanly onto current `master` despite the four commits that have since landed in these files.

On the Windows question — I extracted the Step 10 block verbatim and drove it through each failure mode on Linux: `ln` working, `ln` missing, `ln` copying instead of linking (the MSYS default), a pre-existing real directory, and a stale symlink pointing somewhere else. It reports `linked` / `copy` / `absent` / `skipped-realdir` correctly in all five, always from the read-back rather than from the command having run. Real Windows and the `mklink /J` junction remain unverified here, which matches how the PR labels them.

n0an:
Pushed `eb1c037`. The `<dest>` fix is in, and the rebase surfaced one host question I could not resolve without your call.

## 1. `<dest>` is gone (`eb1c037`)

Named as the final skill directory, so there is one reading left. Template:

```
Somewhere else?  mv ~/.agents/skills/<skill_name> <dest_root>/<skill_name> \
                   && ln -sfn <dest_root>/<skill_name> ~/.claude/skills/<skill_name>
```

The fill rule now states the constraint, shows the form that breaks, and requires the resolved paths be substituted rather than the placeholders printed:

> **`mv` always targets the final skill directory, `<dest_root>/<skill_name>`, never `<dest_root>` itself.**

I ran your exact case both ways against a scratch `$HOME`. Old form: `cat: .../.claude/skills/mybook/SKILL.md: No such file or directory`. New form: link resolves to `~/.copilot/skills/mybook` and the file reads back. All five variants (hub+symlink, host-private, Hermes, project-local, and the relink after a move) carry real paths now.

## 2. The rebase found a host the collapse cannot serve, and it needs your call (`39032e2`)

The branch went conflicting after your last pass. #202 landed Hermes Agent in the same Step 5 table this PR collapses, and Hermes is the first host whose personal root the cross-agent hub cannot reach: `$HERMES_HOME/skills/<category>`, partitioned by category, and it does not scan `~/.agents/skills` for personal installs (only `.agents/skills` project-local).

So "one root every host can reach, natively or by symlink" is no longer true as written, and resolving the conflict by simply taking my side of the table would have sent Hermes users' skills to a root Hermes never reads. That is a regression against current `master`, which installs correctly for them today.

**What I did, deliberately the conservative option:** Hermes keeps its own personal root. The table keeps the Hermes row, a personal Hermes install uses it and takes no symlink, and the "Discoverable by" line names Hermes alone for that destination, with no cross-agent claim, because no other host scans the Hermes root. The intro sentence no longer says one copy serves every host, since with Hermes carved out that is not true. README and the pipeline diagram say the same.

**What I did not do, and want your read on:** give Hermes the same treatment as Claude Code, a link into `$HERMES_HOME/skills/<category>/<skill_name>` with `hermes skills list` as the read-back. It fits the machinery this PR already has, and `hermes skills list` is a genuine read-back rather than an assertion, so it would satisfy the rule the rest of Step 10 runs on. I left it out for two reasons: you said twice you were not looking to expand this PR, and I cannot verify Hermes symlink discovery or the category semantics from here, which puts it in the same "lead, not an instruction" box as the Windows junction. Say the word and it is a follow-up, or say it belongs here and I will add it with the verification wired in.

Also carried `hermes skills trust` and the profile-path warning from #202 through the collapse verbatim, and kept the unknown-host prompt in master's order so `test_hermes_is_in_unknown_host_prompt_and_reload_guidance` keeps passing rather than editing someone else's test to fit mine.

## Gates

`ruff check --select E9,F` clean. `validate_skill.py SKILL.md` exit 0, only the pre-existing body-length soft warning (817 lines, grown in `master`, not by this PR). `pytest tests/ -q`: 622 passed, 5 skipped. Merges cleanly onto current `master`. No new URLs, dependencies, or CI surface; `CHANGELOG.md` untouched.

virgiliojr94 (approved):
The relocation fix and verified symlink reporting address the previous requests. Keep Hermes on its existing personal root. Local suite: 621 passed, 6 skipped; lint and skill validation pass. All five simulated symlink cases report the expected status.

Issue #91, opened 2026-07-31
_chapter_number() misses Markdown-prefixed Latin headings (`## Chapter 1`), inflating chapters_detected in --mode technical


mspindj:
## Summary

`_chapter_number()` does not recognise Markdown-prefixed Latin chapter headings (`## Chapter 1`), so books whose chapters are perfectly well-formed fall through to `_structural_chapter_count()` unnecessarily — and the structural fallback then reports a wildly inflated `chapters_detected`.

This bites specifically in `--mode technical`, because Docling emits headings as Markdown (`## Chapter 1`) rather than as bare lines (`Chapter 1`). So the better extractor is the one that breaks structure detection.

This is adjacent to but distinct from #81. #81 made the structural fallback itself more accurate; this report is about books reaching that fallback at all when they have parseable `Chapter N` headings. It's the same class of gap as #82 (Korean), just for Markdown-prefixed Latin headings.

## Root cause

`_chapter_number()` anchors on `^\s*`, which tolerates leading whitespace but not a Markdown heading prefix:

```python
>>> from book_to_skill.utils import _chapter_number
>>> _chapter_number("Chapter 1")                        # 1
>>> _chapter_number("  Chapter 1")                      # 1
>>> _chapter_number("CHAPTER 1")                        # 1
>>> _chapter_number("## Chapter 1")                     # None   <-- here
>>> _chapter_number("## Chapter 1 Interaction Design")  # None   <-- here
```

There is already a Markdown-aware variant for Chinese headings:

```python
_MD_CN_HEADING = re.compile(rf"^#{{1,6}}\s+第?\s*([{_FW_DIGITS}{_CN_NUM_CLASS}]+)\s*[·、.:：章回卷节篇讲]")
```

but no equivalent for `_EXPLICIT_CHAPTER`. So Markdown-prefixed chapter headings are handled for Chinese and not for English/Spanish/etc.

## Minimal reproduction

No book file needed:

```python
from book_to_skill.utils import detect_structure

md    = "\n".join(f"## Chapter {i}\n## Some Section\nbody\n" for i in range(1, 36))
plain = "\n".join(f"Chapter {i}\nbody\n"                     for i in range(1, 36))

detect_structure(md)     # {'chapters_detected': 36, 'chapter_headings_sample': []}
detect_structure(plain)  # {'chapters_detected': 35, 'chapter_headings_sample': [10 items]}
```

[...]

PR #126, opened 2026-08-09
fix: ToC detection misses Markdown headings; concurrent runs share one workdir

julianobarbosa:
Two independent defects found while benchmarking the converter, plus the workflow changes that follow from them. Happy to split into separate PRs if you'd prefer to take them one at a time.

## 1. ToC detection never fires on Markdown

`_TOC_PATTERN` is `^\s*(table of contents|contents|...)\s*$`, which requires the header alone on its line. Markdown writes it as `## Table of Contents`, so **every** Markdown document reports `ToC: not detected` and emits the "chapter mapping may miss or duplicate sections" warning.

This is the same blind spot #91/#92 fixed for chapter headings — the chapter regex was patched, its ToC sibling was left behind.

The pattern now allows ATX/AsciiDoc markers, emphasis wrappers and a trailing colon (ASCII or fullwidth), while keeping the end-of-line anchor so an inline "the contents of this chapter" still does not match.

Reproducer, before the fix (`book.md` contains a `## Table of Contents` section):

```
$ python3 scripts/extract.py book.md --mode text
   Chapters: 7 detected overall
   ToC     : not detected
   WARN    : No table of contents detected ...
```

## 2. Concurrent runs silently corrupt each other

Every extraction defaults to one shared `<tempdir>/book_skill_work`, so two runs in flight overwrite each other's `full_text.txt`. The failure is silent and worse than a crash: the losing run still writes its own `metadata.json`, which then describes text it did not produce, and nothing in the output indicates the text is wrong.

`claim_workdir()` takes a PID lock and refuses to start on a workdir a live run holds, naming `BOOK_SKILL_WORKDIR` as the way out. Stale locks are reclaimed so a crashed run cannot wedge the next one, and `EPERM` is treated as alive rather than gone so a lock held by another user's process is never stolen.

## 3. SKILL.md workflow changes

These are more opinionated than the two code fixes, so they are the most likely part to want discussion:

- **Destination gate (Step 5).** Step 5 resolved `SKILLS_HOME` to the personal skills root and used it "without asking". That root is auto-loaded every session, so a misfired conversion does not leave a file the user can delete — it leaves a skill that starts firing on their prompts. The gate states the full path and gets agreement before Step 6 creates anything.
- **Chapter count (Step 2.5).** The estimate multiplies a per-chapter budget by `chapters_detected`. A book numbering its sections `1.`/`2.`/`3.` through `pdftotext` yields `0`, which deletes the entire chapter term and collapses the estimate to a flat ~8,500 output tokens for a book of any size. Observed off by roughly 2x on a real 27-page PDF.
- **ToC entries may not resolve (Step 3).** In that same book the ToC listed a `Conclusion` whose only occurrence in the extracted text was the ToC line itself. Step 3 now says to grep-confirm each entry outside the ToC block and never reconstruct a chapter from its title.

## Tests

33 new tests across two new files, kept out of `test_book_to_skill.py` to stay merge-friendly. Suite goes 270 → 304 passing, 1 skipped, no regressions.

One of those tests caught a real bug in my own first attempt: I had written `except OSError` as "process is gone", but `EPERM` means the process exists and merely cannot be signalled, so a root-owned lock would have been wrongly reclaimed.

## Not addressed

`pdftotext` renders `ff`/`ffi` ligatures as `!` (`di!erent`, and `di!` for `diff`). Auto-repair looks unsafe — you cannot distinguish a mangled ligature from a real `!` without a dictionary — so a warning when the signature appears may be the better fix. Left alone here.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -156,6 +156,13 @@ if ! command -v "$PYTHON_BIN" >/dev/null 2>&1; then
   PYTHON_BIN="python"
 fi
 
+# Give this run its own extraction workdir. The default is shared across runs,
+# so two conversions in flight at once overwrite each other's full_text.txt —
+# and the loser still writes a metadata.json describing text it did not produce.
+# The extractor now refuses to start on a workdir another live run holds; this
+# keeps you from hitting that at all. Reuse this same value in Step 10.
+export BOOK_SKILL_WORKDIR="${BOOK_SKILL_WORKDIR:-$(mktemp -d -t book_skill_work-XXXXXX)}"
+
 "$PYTHON_BIN" "$SCRIPT_PATH" $INPUT_PATHS --mode <BOOK_TYPE> --install-missing ask
 ```
 
@@ -163,17 +170,17 @@ Before extraction, the script checks optional Python packages needed for the det
 
 **Tip — preflight the environment:** run `"$PYTHON_BIN" "$SCRIPT_PATH" --check` to print a per-format report of which extractors are installed and the exact command to install whatever is missing, without processing any file. Useful when a user reports a setup or quality problem.
 
-This creates:
-- `<tempdir>/book_skill_work/full_text.txt` — combined extracted text of all sources with clear visually demarcated boundaries.
-- `<tempdir>/book_skill_work/metadata.json` — overall combined size, words, pages, token counts, and a detailed list of individual processed `sources`.
+This creates, inside `$BOOK_SKILL_WORKDIR` (the run prints `Workdir ->`, `Text ->` and `Meta ->` so you can confirm where they landed):
+- `full_text.txt` — combined extracted text of all sources with clear visually demarcated boundaries.
+- `metadata.json` — overall combined size, words, pages, token counts, and a detailed list of individual processed `sources`.
 
-Read `<tempdir>/book_skill_work/metadata.json` to inspect the results.
+Read `$BOOK_SKILL_WORKDIR/metadata.json` to inspect the results.
 
 ---
 
 ## Step 2.5 — Pre-flight cost estimate
 
-Read `<tempdir>/book_skill_work/metadata.json` and present the user with an estimate **before doing any generation**:
+Read `$BOOK_SKILL_WORKDIR/metadata.json` and present the user with an estimate **before doing any generation**:
 
 ```
 📖 Sources detected: <total_sources> source(s)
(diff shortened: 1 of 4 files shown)


virgiliojr94:
> Happy to split into separate PRs if you'd prefer to take them one at a time.

Yes please — taking you up on that. The two defects here are genuinely independent (a regex blind spot in ToC detection, and a concurrency problem in workdir allocation), they carry different risk, and squash-merge turns the PR title into a single changelog entry, so a two-defect PR produces a changelog line that describes half the change.

Splitting also lets the ToC fix land immediately while the workdir isolation gets the closer look it deserves.

On defect 1: the diagnosis matches what I see. `_TOC_PATTERN` requires the header alone on its line, so `## Table of Contents` never matches — the exact blind spot #91/#92 fixed for chapter headings, left behind in the ToC sibling. Note there's a related fix in flight for the CJK side (#112, `目 录` with an inserted space); the two are complementary, not conflicting.

Thanks for benchmarking the converter and reporting what you found rather than working around it.

virgiliojr94:
Heads-up on the red CI here, since it's unrelated to your code: the failing check is **PR description check**, and it's failing legitimately — this PR uses its own headings (`## Problem` / `## Change` / …) rather than the repository template.

The gate does a literal match for `## Summary`, `## Type of change`, `## What it does`, `## Motivation & context`, `## Evidence` and `## Checklist`, needs at least one `- [x]` under Type of change, and needs every Checklist item ticked. It's a required status check, so it has to go green before I can merge.

Editing the PR body is enough — the workflow re-runs on `edited`, no push needed. Your content already covers what the template asks for; it mostly needs to be filed under those headings.

Separately: I've relaxed this repo's fork-PR approval policy, so CI now runs on your pushes without waiting for me to approve it.

virgiliojr94:
Status update so you do not rebase dead work: the concurrent-workdir half you diagnosed here is now fixed on `master` via #184, which credits this PR as prior art. Your reasoning that silent corruption is worse than a crash is what carried the decision.

That leaves the ToC-detection half as the live scope for this PR. #161 also targets ToC detection, so please rebase on current `master` and reduce this to the ToC fix only — I will arbitrate the two on measured behaviour.

virgiliojr94:
Both halves diagnosed in this PR are now resolved on master: concurrent workdir isolation landed via #184, and Markdown ToC detection landed via #161 (both crediting this PR as prior art). Closing with full credit to @julianobarbosa for the original diagnosis and reproductions!

Issue #47, opened 2026-06-15
good first issue: surface extractor errors instead of swallowing them silently


virgiliojr94:
### Context
Every parser in `scripts/extractor/parsers/` follows a fallback chain: try the best extractor, fall back to the next, and so on. The fallback is driven by returning `None`, which is the right design. The problem is the catch-all at the bottom of each extractor:

```python
except ImportError:
    return None        # expected: optional library not installed, fine to stay silent
except Exception:
    return None        # PROBLEM: a corrupt file, an encoding error, a real bug all vanish here
```

When extraction fails, the user gets a generic "could not extract" with no clue why (corrupt PDF? bad encoding? a genuine bug?). The `ImportError` path is a legitimate silent fallback. The broad `except Exception` is not, it hides the actual cause.

There are ~18 of these across the parsers (`pdf.py`, `epub.py`, `docx.py`, `rtf.py`, `calibre.py`, `text.py`).

### The task
Keep the fallback behavior exactly as it is (still return `None` so the next extractor is tried), but make the unexpected error visible. A minimal approach:

```python
import sys
...
except ImportError:
    return None
except Exception as e:
    print(f"  [warn] {extractor_name} failed: {type(e).__name__}: {e}", file=sys.stderr)
    return None
```

### Scope / rules
- Do not change the fallback control flow. Every extractor must still return `None` on failure so the chain continues. This is behavior-preserving plus a warning, nothing more.
- Leave the `except ImportError: return None` branches silent. Missing an optional library is expected, not an error.
- Keep the base install dependency-free (stdlib only). No new packages for this.
- Add or update a test that asserts a deliberately broken input still falls through to `None` and that the chain is not interrupted.

### How we will know it is better
Run any extractor against a deliberately corrupt file: today you get a silent `None`, after this you get one line on stderr naming the extractor and the exception type, while the run still proceeds to the next fallback.

Good first issue for someone comfortable with Python. Happy to review.

addy790:
Hi @virgiliojr94! I'd love to work on this. I'm a beginner contributor and this looks like a great first issue for me. Can I be assigned to this?

virgiliojr94:
Fixed in #50 (thanks @addy790). Unexpected parser exceptions now surface on stderr with the extractor name and exception type, while the fallback chain still returns `None` and continues. Covered by a new test that monkeypatches an extractor to raise and asserts both the `None` return and the stderr warning.

PR #129, opened 2026-08-10
fix(pdf): use reading order for multi-column PDFs instead of -layout

### LINKED ISSUE #128: PDF: `pdftotext -layout` interleaves columns on multi-column documents, and the word-based token estimate hides the 26x size blowup
## Summary

`extract_with_pdftotext()` runs `pdftotext -layout` unconditionally. On a single-column document that is a good default — it keeps table columns aligned. On a **multi-column** document it is wrong twice over, and the second failure is silent.

Fully reproducible with a public-domain document, no copyrighted book needed: **IRS Publication 17 (2025)**, 142 pages, three columns per page, <https://www.irs.gov/pub/irs-pdf/p17.pdf>.

## 1. Reading order — the one that matters

`-layout` emits each physical *line* across all three columns, so consecutive sentences from different columns get interleaved. Searching both extractions for the same phrase:

```
-layout        'standard deduction or if you         (QOF). Taxpayers who made a'

reading order  'Standard deduction amount increased. For 2025, the standard
                deduction amount has been increased for all filers'
```

The first is three columns welded together. Every downstream step — chapter detection, ToC detection, and the agent doing the distillation in Step 3 — reads that.

This is not a cosmetic problem. A skill distilled from interleaved prose will contain sentences that never existed in the book.

## 2. Size

`-layout` pads every line with spaces so columns line up visually. `clean_pdftotext()` strips repeated headers/footers and dehyphenates, but nothing ever collapses that padding.

| extractor | chars | whitespace | ~tokens (chars/4) |
|---|---:|---:|---:|
| `pdftotext -layout` *(current)* | 25,266,249 | **96.8%** | 6,316,562 |
| `pdftotext` (no flag) | 957,757 | 15.8% | 239,439 |
| `pypdf` | 1,002,715 | 15.6% | 250,678 |

The emitted `full_text.txt` is **26× larger** than it needs to be, and 96.8% of it is spaces.

## 3. Why neither is visible

`estimate_tokens()` counts **words**:

```python
return int(len(text.split()) / WORDS_PER_TOKEN)
```

Column padding does not change word count, and word count says nothing at all about ordering. So `metadata.json` reports:

```json
"chars": 25264555,
"words": 165049,
"estimated_tokens": 220065,
"estimated_tokens_human": "~220K"
```

`chars` is right there and is 25M, but the CLI surfaces the token figure — so the run reports **~220K tokens for a file that costs ~6.3M to read**. The number that gets checked and the number that gets paid are different numbers.

The same run also reported `Chapters: 3 detected` for a publication with dozens, and correctly warned that no ToC was found. I suspect both are downstream of the interleaving rather than separate bugs, though I have not proven that.

## Repro

```bash
curl -sLO https://www.irs.gov/pub/irs-pdf/p17.pdf
python3 scripts/extract.py p17.pdf
# metadata.json: estimated_tokens ~220K, chars 25,264,555, chapters_detected 3

pdftotext -layout p17.pdf - | grep -o 'standard deduction[^\n]\{0,60\}' | head -1
pdftotext          p17.pdf - | grep -o 'Standard deduction[^\n]\{0,60\}' | head -1
```

## Suggested fix

**Detect columns and pick the mode.** A crude signal works well: under `-layout`, count substantive lines containing a wide internal run of spaces (the gutter). On Pub 17 that is 85% of sampled lines; on a single-column PDF it is near zero.

```python
GUTTER = re.compile(r"\S[ ]{4,}\S")

def looks_multicolumn(layout_text, threshold=0.35):
    lines = [ln for ln in layout_text.splitlines() if len(ln.strip()) > 40]
    if not lines:
        return False
    return sum(1 for ln in lines[:4000] if GUTTER.search(ln)) / len(lines[:4000]) >= threshold
```

Then use plain `pdftotext` (reading order) when multi-column, keep `-layout` otherwise. Both extractions are cheap — the whole thing took 11.7s on 142 pages — so running both and choosing is affordable.

Smaller changes that would each help on their own:

1. **Estimate tokens from characters, not words,** or report both. The word-based figure is defensible as a proxy for content volume but is not what the run will cost. This alone would have made the 26× blowup visible at conversion time.
2. **Collapse the padding after `clean_pdftotext()`** for the cases where `-layout` is still the right choice. Collapsing runs of 2+ spaces to one took Pub 17 from 25,264,555 to 970,243 chars with all 165,049 words intact — but note this makes interleaved text *smaller*, not *correct*, so it is not a substitute for fix (1) above.
3. **Warn when `-layout` output looks multi-column,** even without changing the mode. Consistent with the reasoning in #47 — surface it rather than swallow it.

Happy to open a PR for the detection approach if the shape looks right to you. I have it working locally against this document.

## Environment

- book-to-skill `442aaaa2d21dbe5ae0f7c6a396585ef3a9d2534e` (v1.3.0)
- Windows 11, Python 3.12
- poppler `pdftotext`, `pypdf` and `pdfminer.six` all installed; `--check` reports the PDF chain ready
- `--mode text` (default); docling not installed, so `--mode technical` was not exercised

Bwoodnva:
Fixes #128.

## Problem

`extract_with_pdftotext()` runs `pdftotext -layout` unconditionally. That is the right default for a single-column document — it keeps table columns aligned. On a **multi-column** document every emitted line spans all columns, so consecutive sentences from different columns interleave.

Measured on **IRS Publication 17** (142 pages, three columns, public domain, <https://www.irs.gov/pub/irs-pdf/p17.pdf>), searching both extractions for the same phrase:

```
-layout        'standard deduction or if you        (QOF). Taxpayers who made a'

reading order  'Standard deduction amount increased. For 2025, the standard
                deduction amount has been increased for all filers'
```

The first is three columns welded together — and it is what chapter detection, ToC detection and the Step 3 distillation all read. A skill built from it can contain sentences that never existed in the book.

`-layout` also pads every line to preserve horizontal position, and `clean_pdftotext()` never collapses it:

| | chars | whitespace |
|---|---:|---:|
| `-layout` | 25,266,249 | 96.8% |
| reading order | 957,007 | 15.8% |

26× larger, for text that is also wrong. That blowup is invisible today because `estimate_tokens()` counts words, and padding does not change word count — so `metadata.json` reports `~220K` tokens for a file that costs ~6.3M to read.

## Change

Detect the layout and pick the mode.

Detection is a structural signal rather than layout analysis: under `-layout`, a wide run of spaces between two runs of real text means two blocks sat side by side.

```python
_PDF_GUTTER = re.compile(r"\S {4,}\S")
_MULTICOLUMN_THRESHOLD = 0.35
_MIN_GUTTER_LINE = 40      # headings and page furniture are not evidence
_GUTTER_SAMPLE = 4000
```

Publication 17 scores **0.85**. A single-column document scores near zero.

**`-layout` still runs first**, so the single-column majority is completely unchanged and pays no extra subprocess. The reading-order pass runs only when detection fires.

If that second pass fails, the `-layout` text is still returned — interleaved text beats no text — but with a warning rather than silently. That follows the reasoning in #47 about surfacing rather than swallowing.

Stdlib only; no new dependencies.

## Verification

End to end on Publication 17: detection fires, output drops to 957,007 chars, and the phrase above comes back intact.

```
  [info] multi-column layout detected - using reading order instead of -layout
  chars=957,007  whitespace=15.8%  ~tokens=239,251
  phrase: Standard deduction amount increased. For 2025, the standard deduction
          amount has been increased for all filers. The amounts are:
```

**11 new tests** in `tests/test_pdf_multicolumn.py`:

- detection on single-column vs two-column text
- empty and whitespace-only input
- short-lines-only input (headings and page numbers must not trigger it)
- a table inside an otherwise single-column book must not flip the whole document
- the single-column path still runs **exactly one** subprocess, with `-layout`
- the multi-column path re-runs without it
- a failed second pass warns and keeps the `-layout` text
- a missing `pdftotext` binary still returns `None`, so the fallback chain continues
- the gutter regex matches a real Publication 17 line and not ordinary prose

`pytest -q`: **282 passed**. `ruff check`: clean.

## Notes

- The same run reported `Chapters: 3 detected` on a publication with dozens. I suspect that is downstream of the interleaving, but I have not proven it and this PR does not claim to fix it.
- `estimate_tokens()` counting words rather than characters is the reason neither failure is visible in `metadata.json`. Left alone here to keep this PR to one change — happy to open a separate one if you want it.
- Not addressed: collapsing `-layout` padding for documents where `-layout` remains the correct choice. Collapsing runs of 2+ spaces took Publication 17 from 25,264,555 to 970,243 chars with all 165,049 words intact, but that makes interleaved text *smaller*, not *correct*, so it is a separate concern from this fix.

### DIFF
--- a/book_to_skill/parsers/pdf.py
+++ b/book_to_skill/parsers/pdf.py
@@ -10,6 +10,41 @@ from collections import Counter
 _PDF_PAGE_NUM = re.compile(r"^\s*(\d{1,4}|[ivxlcdm]{1,7})\s*$", re.IGNORECASE)
 _PDF_HYPHEN_WRAP = re.compile(r"(\w)-\n(\w)")
 
+# A wide run of spaces *inside* a line of text is the gutter between columns.
+# Four is above the widest inter-word justification pdftotext emits and below
+# the narrowest real gutter.
+_PDF_GUTTER = re.compile(r"\S {4,}\S")
+
+# Fraction of substantive lines that must show a gutter before the document is
+# treated as multi-column. A single-column PDF scores near zero; a
+# three-column government publication scores ~0.85.
+_MULTICOLUMN_THRESHOLD = 0.35
+
+# Short lines are headings, page furniture and list stubs; they pick up wide
+# gaps by accident. Only lines long enough to span a column are evidence.
+_MIN_GUTTER_LINE = 40
+_GUTTER_SAMPLE = 4000
+
+
+def looks_multicolumn(layout_text: str) -> bool:
+    """Does ``pdftotext -layout`` output come from a multi-column document?
+
+    ``-layout`` preserves horizontal position, which is what keeps table
+    columns aligned — and is exactly what makes a multi-column page unusable:
+    every emitted line spans all columns, so consecutive sentences from
+    different columns interleave.
+
+    This is a structural signal, not layout analysis: two runs of real text
+    separated by a wide space run means two blocks sat side by side.
+    """
+    lines = [ln for ln in layout_text.splitlines()
+             if len(ln.strip()) > _MIN_GUTTER_LINE]
+    if not lines:
+        return False
+    sample = lines[:_GUTTER_SAMPLE]
+    hits = sum(1 for ln in sample if _PDF_GUTTER.search(ln))
+    return hits / len(sample) >= _MULTICOLUMN_THRESHOLD
+
 
 def clean_pdftotext(text: str) -> str:
     """Clean pdftotext '-layout' output (pages are form-feed delimited): drop
@@ -47,18 +82,58 @@ def clean_pdftotext(text: str) -> str:
     return _PDF_HYPHEN_WRAP.sub(r"\1\2", text)
 
 
+def _pdftotext(pdf_path: str, *args: str) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        ["pdftotext", *args, pdf_path, "-"],
+        capture_output=True, text=True, timeout=120,
+        encoding="utf-8", errors="replace",
+    )
+
+
 def extract_with_pdftotext(pdf_path: str) -> str | None:
+    """Extract a PDF, choosing ``-layout`` only when it will not scramble text.
+
+    ``-layout`` is the right default for a single-column document: it keeps
+    table columns aligned. On a multi-column document it destroys reading
+    order, because each emitted line spans every column. Measured on IRS
+    Publication 17 (142 pages, three columns), searching both outputs for the
+    same phrase:
+
+        -layout        'standard deduction or if you    (QOF). Taxpayers who made a'
+        reading order  'Standard deduction amount increased. For 2025, the
+                        standard deduction amount has been increased for all filers'
+
+    The first is three columns welded together, and it is what every later
+    step reads. ``-layout`` also pads each line to preserve position, which on
+    that document produced 25,266,249 characters at 96.8% whitespace against
+    957,757 in reading order — 26x larger for text that is also wrong.
+
+    So: run ``-layout`` first (unchanged for the single-column majority), and
+    only re-run without it when the output looks multi-column. The second
+    subprocess is paid solely by documents that need it.
+    """
     if not shutil.which("pdftotext"):
         return None
     try:
         pdf_path = os.path.abspath(pdf_path)
-        result = subprocess.run(
-            ["pdftotext", "-layout", pdf_path, "-"],
-            capture_output=True, text=True, timeout=120,
-            encoding="utf-8", errors="replace",
-        )
-        if result.returncode == 0 and result.stdout.strip():
-            return clean_pdftotext(result.stdout)
+        result = _pdftotext(pdf_path, "-layout")
+        if result.returncode != 0 or not result.stdout.strip():
+            return None
+
+        if looks_multicolumn(result.stdout):
+            reading_order = _pdftotext(pdf_path)
+            if reading_order.returncode == 0 and reading_order.stdout.strip():
+                # ASCII only: this goes to stderr from the parser, which does
+                # not get cli.py's UTF-8 reconfigure on a legacy Windows console.
+                print("  [info] multi-column layout detected - using reading "
+                      "order instead of -layout", file=sys.stderr)
+                return clean_pdftotext(reading_order.stdout)
+            # Falling back to -layout output is still better than nothing;
+            # say so rather than silently returning scrambled text.
+            print("  [warn] multi-column layout detected but the reading-order "
+                  "pass failed; text may be interleaved", file=sys.stderr)
+
+        return clean_pdftotext(result.stdout)
     except Exception as e:
         print(f"  [warn] extract_with_pdftotext failed: {type(e).__name__}: {e}", file=sys.stderr)
     return None
--- a/tests/test_pdf_multicolumn.py
+++ b/tests/test_pdf_multicolumn.py
@@ -1,0 +1,140 @@
+"""`pdftotext -layout` must not be used on multi-column documents.
+
+`-layout` preserves horizontal position, which keeps table columns aligned on a
+single-column page and destroys reading order on a multi-column one: every
+emitted line spans all columns, so consecutive sentences from different columns
+interleave.
+
+Measured on IRS Publication 17 (142 pages, three columns per page), searching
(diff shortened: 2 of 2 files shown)


virgiliojr94:
This is a well-argued PR and the underlying problem (#128) is real — `-layout` on a three-column document interleaves sentences, and the failure is silent, which is the class of bug I care most about. Thank you for reproducing it on a public document instead of a copyrighted book; that made it possible to check independently.

I'm not merging it as-is, and the reason is the detector rather than the fix.

`looks_multicolumn()` decides globally, from a 4000-line sample, whether the whole document is multi-column, using `\S {4,}\S` at a 0.35 threshold. My concern is the false positive: `-layout` exists in this project specifically to keep **tables** aligned, and a table row is exactly "text, wide space run, text" on a long line. A reference book that is single-column prose but table-dense — the technical books this tool is most used on — could cross 0.35 and silently lose the alignment `--mode technical` is there to preserve. That trades a visible failure on multi-column PDFs for an invisible one on table-heavy PDFs.

Two things would change my mind, either one:

1. **Measurement against the false positive.** Run the detector over a set of single-column, table-dense technical PDFs and report the scores. If they cluster well below 0.35 with real margin, the threshold is defensible and I'll take it.
2. **Per-page instead of per-document.** Column layout is a property of a page, not a book — front matter, full-width tables and appendices routinely differ from the body. Deciding per page removes the need for a global threshold to be right about a whole document, and the failure mode shrinks to one page.

The `_pdftotext()` helper and the reading-order fallback themselves look right to me. It's specifically the "is this document multi-column" call that needs to be harder to get wrong.

virgiliojr94:
Heads-up on the red CI here, since it's unrelated to your code: the failing check is **PR description check**, and it's failing legitimately — this PR uses its own headings (`## Problem` / `## Change` / …) rather than the repository template.

The gate does a literal match for `## Summary`, `## Type of change`, `## What it does`, `## Motivation & context`, `## Evidence` and `## Checklist`, needs at least one `- [x]` under Type of change, and needs every Checklist item ticked. It's a required status check, so it has to go green before I can merge.

Editing the PR body is enough — the workflow re-runs on `edited`, no push needed. Your content already covers what the template asks for; it mostly needs to be filed under those headings.

Separately: I've relaxed this repo's fork-PR approval policy, so CI now runs on your pushes without waiting for me to approve it.

virgiliojr94 (dismissed):
✅ **Approved** — this is the cleanest implementation of multi-column PDF detection and arrived first (Aug 10). The heuristic (`looks_multicolumn()` sampling gutter lines) correctly identifies multi-column layouts while preserving `-layout` for single-column documents, and the two-pass approach ensures no extra cost for the common case.

**Cluster context:** two PRs (#129, #166) independently fixed #128. This one wins on:
- **Timing**: opened Aug 10, five days before #166 (Aug 15)
- **State**: OPEN vs DRAFT
- **Correctness**: 11 comprehensive tests covering detection edge cases and mode selection
- **Design**: stdlib-only, no new dependencies, minimal changes to the hot path

I'll merge this shortly and close #166 as a duplicate with credit. Thanks for the thorough fix and detailed evidence!

cc @aryansk (author of #166)

virgiliojr94:
📝 **PR description needs update** — the automated check is failing because the PR body doesn't match the required template sections. This is an easy fix; no code changes needed.

**Missing sections (from .github/PULL_REQUEST_TEMPLATE.md):**
- `## Summary` (you have "## Problem" — rename it)
- `## Type of change` (tick at least one box: 🐛/✨/⚡/♻️/📝/🔒/🧹/💥)
- `## What it does` (you have "## Change" — rename it)
- `## Motivation & context` (you have parts in "## Problem" — move context here)
- `## Evidence` (you have "## Verification" — rename it)
- `## Checklist` (required — all items must be ticked [x])

**What to do:**
1. Edit the PR description
2. Rename your existing sections to match the template
3. Add the Type of change checkboxes and Checklist
4. Push the description update (no code change needed)

Once updated, the "PR description check" will pass and I can merge this. Thanks!

virgiliojr94:
Thanks for tackling #128. The diff is focused and the reported IRS Pub. 17 result is useful.\n\nBefore I can arbitrate this cluster, please add a reproducible measurement path: identify the exact public fixture/version and command(s), then show the detector result plus `-layout` vs reading-order character/whitespace counts. The new regression test also needs to fail on `master` and pass on this head.\n\n#166 solves the same root cause, so I will select the measured implementation rather than merge both.

PR #134, opened 2026-08-10
fix(epub): report unextracted source images

### LINKED ISSUE #127: EPUB extraction silently discards all images, losing figure-borne content
# EPUB extraction silently discards all images, losing figure-borne content

## Summary

The EPUB parsers extract text only. Every image in the archive is dropped without a warning, so books whose content lives in figures — checklists, comparison matrices, quadrant diagrams, decision tables — produce a skill that is confidently incomplete. Nothing in the output signals that anything was lost.

Converting *Systems Architecting: A Business Perspective* (Gerrit Muller, CRC Press 2011) surfaced this: **226 images, 7.8 MB — 93% of the EPUB.** The text is 0.6 MB.

```
images in EPUB : 226  (7.8 MB)
text documents : 18   (0.6 MB)
```

The generated skill cited figures like `Figure 3.10 The key-driver graph of a motorway management system` while containing none of what the figure shows. Captions survive as orphan lines in `full_text.txt`; their payload does not.

## Root cause

Two independent places, both in `book_to_skill/parsers/epub.py`:

```python
# extract_with_ebooklib
for item in book.get_items_of_type(ebooklib.ITEM_DOCUMENT):
    ...
```

```python
# extract_with_zipfile
html_files = spine_order or sorted(
    n for n in names if n.endswith((".html", ".xhtml"))
)
```

`ITEM_IMAGE` is never requested; non-HTML members are never read.

Underneath that is an architectural constraint rather than an oversight: every parser returns `str | None`, and `utils.py` funnels the result into a single `full_text.txt`. **There is no place in the output contract for a binary.** Fixing the two loops is not sufficient on its own — the pipeline needs an asset sink.

## Why this is worse than it first appears

1. **It is silent.** A missing extractor for RTF or MOBI prints a warning. This prints nothing, so a user has no signal to check.
2. **It is invisible downstream.** The distillation step sees caption text without images and cannot tell the difference between "figure was decorative" and "figure was the content".
3. **It is inverted relative to the docs.** `SKILL.md` Step 1.5 offers a *technical* mode promising "tables, code blocks, formulas preserved as markdown". For EPUB that promise cannot be kept: `extract_with_docling` lives in `parsers/pdf.py` and is only called on the `.pdf` branch, so `--mode technical` is a no-op for EPUB. A user selecting technical mode for a figure-heavy EPUB gets prose-only extraction and a mode confirmation message implying otherwise.
4. **PDF is inconsistent with EPUB.** Docling on the PDF path does export figures. The same book in two formats yields materially different skills.

## Reproduce

```bash
python3 scripts/extract.py some-figure-heavy-book.epub --mode technical
grep -c "^Figure " /tmp/book_skill_work/full_text.txt   # captions present
# no image is written anywhere; no warning is emitted
```

## Recovery is tractable

The figures are recoverable after the fact, which suggests the fix is not expensive. Images carry no figure number of their own (`Image000NN.jpg`), but in document order each `` is followed by its caption:

```
IMG Image00008.jpg → CAP Figure 1.2
IMG Image00009.jpg → CAP Figure 1.3
IMG Image00010.jpg → CAP Figure 1.4
```

Walking the OPF spine and assigning each image the next `Figure N.M` caption recovered **197 of 204 numbered figures with 0 corrupt files and 0 unresolved captions**; the 8 skipped were sub-1 KB spacers. That is a positional heuristic, not a parse — an in-tree fix could do better by reading the surrounding `<figure>`/`<figcaption>` or `alt` attributes where present.

## Suggested fix

Smallest change that removes the silent failure, in increasing order of effort:

1. **Warn.** If an EPUB (or DOCX, which has the same shape) contains images and none are extracted, print a count to stderr and record it in `metadata.json` as e.g. `"images_dropped": 226`. Cheap, and it converts a silent loss into an informed decision. **Worth doing even if nothing else is.**
2. **Correct the mode messaging.** Step 1.5's technical-mode confirmation should not promise structure preservation for formats that never reach Docling. Either say so, or route EPUB through a structure-aware path.
3. **Add an asset sink.** Let parsers optionally return `(text, assets)` and have `utils.py` write assets to `<workdir>/assets/`, with `full_text.txt` referencing them as `Figure 3.10`. Step 7 could then copy the referenced subset into `<skill>/assets/` and reference them from chapter files.
4. **Extend to DOCX and HTML,** which have the same text-only shape.

Option 1 alone would have changed the outcome here: the loss would have been visible at conversion time instead of discovered later by reading the generated skill against the book.

## Environment

- book-to-skill, `parsers/epub.py` (both `extract_with_ebooklib` and `extract_with_zipfile` paths)
- `ebooklib` not installed, so the stdlib `zipfile` fallback ran — **the ebooklib path has the same defect**, so installing it does not help
- macOS, Python 3.13

hydraxman:
## Summary (Required)

EPUB extraction now reports how many source images were omitted, records the count in per-source and aggregate metadata, and carries material image loss into generated-skill guidance.

## Type of change (Required)

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)

- **Fixes:** EPUB images were silently omitted by the text-only extractors, leaving users and generated skills unaware of missing figure content.
- **Improves:** Every EPUB image member is counted without adding dependencies; raw counts remain available in metadata for every EPUB.
- **Adds:** A warning plus preflight and Scope & Limits guidance when an EPUB contains more than five images. Cover-only and other small image sets stay quiet.

Figure-heavy EPUBs can yield confidently incomplete skills because captions survive while their images do not. This keeps image interpretation out of scope while making material extraction loss visible without warning on ordinary cover-only novels.

Fixes #127

The EPUB parser counts image archive members by their standard image extensions. The extraction result exposes that count and batch metadata aggregates it unconditionally. User-facing warnings and generated-skill disclosure are gated at more than five images so common cover and decorative assets do not create noise.

## Evidence (Required)

- **Tests:** Added focused coverage for material image-loss warnings, a quiet cover-only EPUB, and per-source/aggregate metadata.
- **Before / after:** The cover-only regression fails on the previous PR head because a one-image EPUB emits the warning; it passes after applying the materiality threshold.

```
Before review fix: 1 failed, 2 passed (cover-only EPUB emitted warning)
After review fix:  3 passed (focused image-reporting tests)
Full:              479 passed, 1 skipped
Ruff:              All checks passed!
Skill:             valid; one pre-existing soft length warning
```

```text
[warn] figures.epub contains 6 image(s); their content is not extracted
```

```json
{
  "images_dropped": 6,
  "sources": [
    {
      "filename": "figures.epub",
      "images_dropped": 6
    }
  ]
}
```

- [x] `ruff check .` is clean

No existing extraction return type or text output changes. Non-EPUB sources expose `images_dropped: null`; batch totals remain numeric.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -178,7 +178,7 @@ Read `<tempdir>/book_skill_work/metadata.json` and present the user with an esti
 ```
 📖 Sources detected: <total_sources> source(s)
 <list each source filename and format from the sources metadata list>
-<if images_dropped > 0: warn that N source images were not read>
+<if images_dropped > 5: warn that N source images were not read>
 📄 Combined Pages/Sections: ~<N> | Words: ~<N> | Total tokens: ~<N>K
 
 💰 Estimated token cost (Full Conversion / Update):
@@ -516,7 +516,7 @@ the relevant chapter file before answering.
 This skill covers the book content only. For hands-on implementation in your codebase,
 combine with project-specific tools. For topics beyond this book, check related skills
 or ask the agent directly.
-<if images_dropped > 0: state that N source images were not read>
+<if images_dropped > 5: state that N source images were not read>
 ```
 
 ---
--- a/SKILL.md
+++ b/SKILL.md
@@ -165,7 +165,7 @@ Before extraction, the script checks optional Python packages needed for the det
 
 This creates:
 - `<tempdir>/book_skill_work/full_text.txt` — combined extracted text of all sources with clear visually demarcated boundaries.
-- `<tempdir>/book_skill_work/metadata.json` — overall combined size, words, pages, token counts, and a detailed list of individual processed `sources`.
+- `<tempdir>/book_skill_work/metadata.json` — overall combined size, words, pages, token counts, dropped EPUB image counts, and a detailed list of individual processed `sources`.
 
 Read `<tempdir>/book_skill_work/metadata.json` to inspect the results.
 
@@ -178,6 +178,7 @@ Read `<tempdir>/book_skill_work/metadata.json` and present the user with an esti
 ```
 📖 Sources detected: <total_sources> source(s)
 <list each source filename and format from the sources metadata list>
+<if images_dropped > 0: warn that N source images were not read>
 📄 Combined Pages/Sections: ~<N> | Words: ~<N> | Total tokens: ~<N>K
 
 💰 Estimated token cost (Full Conversion / Update):
@@ -515,6 +516,7 @@ the relevant chapter file before answering.
 This skill covers the book content only. For hands-on implementation in your codebase,
 combine with project-specific tools. For topics beyond this book, check related skills
 or ask the agent directly.
+<if images_dropped > 0: state that N source images were not read>
 ```
 
 ---
--- a/book_to_skill/parsers/epub.py
+++ b/book_to_skill/parsers/epub.py
@@ -7,6 +7,20 @@ import sys
 from book_to_skill.parsers.html import _HTMLTextExtractor
 
 
+_IMAGE_EXTENSIONS = (
+    ".avif",
+    ".bmp",
+    ".gif",
+    ".jpeg",
+    ".jpg",
+    ".png",
+    ".svg",
+    ".tif",
+    ".tiff",
+    ".webp",
+)
+
+
 def extract_with_ebooklib(epub_path: str) -> str | None:
     try:
         import ebooklib
@@ -121,5 +135,17 @@ def count_epub_chapters(epub_path: str) -> int:
             opf_text = zf.read(opf_path).decode("utf-8", errors="replace")
             return len(re.findall(r'<itemref\b', opf_text))
     except Exception:
+        return 0
+
+
+def count_epub_images(epub_path: str) -> int:
+    """Count image members whose content the text-only EPUB parsers omit."""
+    try:
+        with zipfile.ZipFile(epub_path) as zf:
+            return sum(
+                not info.is_dir() and info.filename.lower().endswith(_IMAGE_EXTENSIONS)
+                for info in zf.infolist()
+            )
+    except (OSError, zipfile.BadZipFile):
         return 0
 
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -46,6 +46,7 @@ from book_to_skill.parsers.epub import (
     extract_with_ebooklib,
     extract_with_zipfile,
     count_epub_chapters,
+    count_epub_images,
 )
 from book_to_skill.sanitize import sanitize_extracted_text
 
@@ -635,6 +636,7 @@ def extract_single_file(input_path: Path, extraction_mode: str, install_mode: st
     method = ""
     pages = 0
     pages_label = "sections"
+    images_dropped = None
     
     if ext == ".epub":
         print(f"Extracting EPUB: {input_str}")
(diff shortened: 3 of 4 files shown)


virgiliojr94:
Thanks for taking this — it does the three things I asked for in #127 (count, warn, say it in Scope & Limits) without pretending to read the figures, and it stays dependency-free. Checked out locally: 351 passed, `ruff check .` clean, `validate_skill.py` fine.

One change before I merge, and it is about signal rather than correctness.

`count_epub_images()` counts every archive member with an image extension. That includes the **cover** — and virtually every EPUB in existence has one:

```python
# minimal EPUB: cover.jpg, one real figure, one unreferenced leftover, a font
count_epub_images(...) -> 4   # font excluded, cover and leftover counted
```

So after this lands, a plain prose novel with nothing but a cover prints `[warn] book.epub contains 1 image(s); their content is not extracted` and the generated skill carries a Scope & Limits line about unread figures. #127 was about *Systems Architecting* — 226 images, 93% of the archive. A warning that fires on every EPUB is a warning nobody reads, and the Scope & Limits line stops meaning anything when it is present in every generated skill.

What I want:

- Keep `images_dropped` in `metadata.json` unconditionally — the raw count is data, and it belongs there for every EPUB, cover included.
- Gate the **warning** and the **Scope & Limits** line on the count being material. Simplest thing that works: a small floor (say `> 5`), or better, images relative to text (a book with more images than spine items is figure-borne; a novel with one is not). Your call which — pick one, state the threshold in the code comment, and cover the "cover-only EPUB stays quiet" case in a test.

Two smaller notes, neither blocking:

- The count includes unreferenced archive members too. That is fine by me — the warning says "contains", which is literally true, and reading the OPF manifest to be exact is more machinery than this earns.
- Aggregate `images_dropped: 0` for a PDF-only run reads as "nothing was lost", but a PDF without docling drops figures too, and so does the Calibre path for MOBI/AZW3. Not this PR's job to fix; just don't let the field grow into a claim it can't back. `null` when no source is an EPUB would be more honest if it's cheap.

Sort the threshold and this goes in.

hydraxman:
Addressed the warning-signal feedback in `44ed3eb`.

The raw `images_dropped` count is still recorded for every EPUB, but user-facing warnings and generated Scope & Limits guidance now require more than five images. A new cover-only regression failed on the previous head and now stays quiet while preserving `images_dropped: 1`.

Validation after rebasing onto current `master`:
- `python -m pytest -q tests/test_epub_image_reporting.py` — 3 passed
- `python -m pytest -q` — 479 passed, 1 skipped
- `ruff check .` — passed
- `python tools/validate_skill.py SKILL.md` — passed with the existing soft length warning
- `git diff upstream/master...HEAD --check` — passed

virgiliojr94 (approved):
✅ **Approved** — this is the cleanest implementation of EPUB image reporting and arrived first (Aug 10). The extension-based count (`count_epub_images()`) is stdlib-only, the threshold (>5 images) keeps cover-only novels quiet, and the metadata + warning plumbing makes the loss visible without noise.

**Cluster context:** two PRs (#134, #167) independently fixed #127. This one wins on:
- **Timing**: opened Aug 10, five days before #167 (Aug 15)
- **State**: OPEN with all checks passing vs incomplete description
- **Design**: simple extension scan, sensible threshold, complete metadata
- **Correctness**: 3 focused tests covering the warning, quiet cover-only, and metadata

I'll merge this shortly and close #167 as a duplicate with credit. Thanks for the thorough fix!

cc @aryansk (author of #167)

PR #135, opened 2026-08-11
fix(extractor): stop an unbalanced code fence swallowing later headings

Hotragn:
## Summary

`_structural_chapter_count` tracked fenced code blocks by flipping a boolean on every ``` or `~~~` line. A fence that never closed left the scanner "inside a code block" for the rest of the file, so **every heading after it was dropped**. On an 8-section handbook with one missing closing fence, 5 of 8 sections disappeared.

## Type of change

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does

- **Fixes:** an unterminated code fence silently truncated the structural heading scan at that point.
- **Also fixes:** the toggle ignored *which* fence character opened the block, so a ` ``` ` block could be "closed" by an unrelated `~~~` line — and a stray `~~~` cost a section on its own.

## Motivation & context

Closes n/a — found while auditing the structural fallback; no existing issue.

The blast radius is well targeted at the wrong people: `_structural_chapter_count` only runs when numeric `Chapter N` detection finds nothing, i.e. for **Markdown/AsciiDoc technical docs** — precisely the sources most likely to be full of fenced code blocks.

Two realistic ways a fence ends up unbalanced:

1. **Extraction drops one.** A PDF/EPUB→text pass can lose a line, and a fence is a single line.
2. **The book contains a stray one.** A book *about* Markdown demonstrates fence syntax; a partial fence in prose is normal content.

The failure is silent — extraction reports success, `chapters_detected` is just smaller, and Step 3 plans from a book that appears to have a third of its real structure.

## How it works

Fences are resolved in a pre-pass, `_closed_fence_line_numbers()`, which records only the line indices inside a **closed** block. An unterminated opener is left as ordinary text.

```python
if opener is None:
    opener = (marker[0], index)
elif marker[0] == opener[0]:          # same fence character, per CommonMark
    inside.update(range(opener[1], index + 1))
    opener = None
```

The trade is deliberate and asymmetric: an unclosed fence now means a handful of code lines get scanned as prose, which at worst adds a spurious heading candidate that the existing digit-led / all-punctuation guards mostly reject. The old behaviour lost most of a book's structure. Same trade I made for unterminated RTF destination groups in #109, which you merged — malformed input should degrade, not truncate.

Requiring the closing fence to match its opener's character is straight CommonMark and fixes the `~~~` case as a side effect.

Rejected alternative: keeping the live toggle and just resetting `in_fence = False` at EOF. That does not help — the headings have already been skipped by then; the state has to be known *before* the scan, which is what the pre-pass gives.

## Evidence

**Before / after** on an 8-section Markdown handbook, where section 3 shows a fenced block whose closer was lost:

```
                        before   after
balanced fences            8       8
ONE unclosed fence         3       8      <- 5 of 8 sections were lost
stray "~~~", no closer     2       3
```

`detect_structure` on the broken document went from `{'chapters_detected': 3, …}` to `{'chapters_detected': 8, …}`.

**Tests** — new `tests/test_unbalanced_code_fence.py`, 16 cases in four groups:

- `TestUnbalancedFenceDoesNotSwallowHeadings` — unclosed backtick fence, equality with the balanced document, unclosed `~~~`, and a trailing lone fence marker.
- `TestBalancedFencesStillSuppressHeadings` — **the guard this PR must not break**: ATX and setext headings inside closed fences are still ignored, across single, multiple and tilde fences.
- `TestFenceCharacterMustMatch` — backtick not closed by tilde, matching pairs detected, longer ` ```` ` markers, indented fences, no-fence documents.
- `TestExistingBehaviourPreserved` — digit-led titles still rejected, setext headings still counted, no-heading documents still 0.

```
$ python -m pytest tests/ -q
363 passed, 3 skipped in 4.85s

$ ruff check .
All checks passed!
```

## Screenshots / output

The before/after table above is the output — this is a count, so there is nothing visual to capture.

## Checklist

- [x] `ruff check .` is clean

## Notes for reviewers

**Honest note on RED evidence.** The test file imports the new `_closed_fence_line_numbers`, so running it against unmodified `master` raises an `ImportError` rather than clean assertion failures. The behavioural before/after above is therefore the evidence, not a pytest RED run. If you would rather have a fails-by-assertion test, the four `TestUnbalancedFenceDoesNotSwallowHeadings` cases only need `_structural_chapter_count` and would fail properly on their own — happy to split the direct-unit tests into a follow-up.

**One conscious simplification:** CommonMark also requires the closing fence to be *at least as long* as the opener, so ` ```` ` opens a block that a 3-backtick line does not close. I did not implement that — it would change which lines count as fenced in documents that nest fences, and this PR is about not losing headings rather than about full CommonMark conformance. Say the word if you want it and I will add the length check plus tests.

**Not in scope but adjacent**, happy to send separately: `estimate_tokens` counts BMP CJK but not the supplementary plane (Ext B–G, U+20000+), so a text of rare/classical ideographs estimates at roughly 1/1000 of its real token cost — the same class of undercount #103 fixed for the BMP.

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -181,6 +181,40 @@ _ATX_HEADING = re.compile(r"^(#{1,6}|={1,6})\s+(.+?)\s*#*$")
 _SETEXT_UNDERLINE = re.compile(r"^(={2,}|-{2,})$")
 
 
+# Opening or closing line of a fenced code block: three or more backticks or
+# tildes. The captured marker lets the closer be matched to its opener.
+_CODE_FENCE = re.compile(r"^(`{3,}|~{3,})")
+
+
+def _closed_fence_line_numbers(lines: list[str]) -> set[int]:
+    """Line indices inside a fenced code block that is actually CLOSED.
+
+    A fence that never closes is treated as ordinary text rather than swallowing
+    everything after it. Extraction routinely loses a closing fence, and a book
+    about Markdown can simply contain a stray one — and the old live-toggling
+    scan then dropped every heading from that point to the end of the document.
+    Counting a handful of code lines as prose is a far cheaper mistake than
+    losing most of a book's structure.
+
+    The closing fence must use the SAME character as its opener, per CommonMark,
+    so a "```" block is no longer terminated by an unrelated "~~~" line.
+    """
+    inside: set[int] = set()
+    opener: tuple[str, int] | None = None
+    for index, line in enumerate(lines):
+        match = _CODE_FENCE.match(line.strip())
+        if not match:
+            continue
+        marker = match.group(1)
+        if opener is None:
+            opener = (marker[0], index)
+        elif marker[0] == opener[0]:
+            # Include both fence marker lines themselves.
+            inside.update(range(opener[1], index + 1))
+            opener = None
+    return inside
+
+
 def _structural_chapter_count(text: str) -> int:
     """Count chapter-like structural headings in Markdown/AsciiDoc/RST sources.
 
@@ -198,17 +232,14 @@ def _structural_chapter_count(text: str) -> int:
     not match).
     """
     levels: dict[int, set[str]] = {}
-    in_fence = False
+    lines = text.splitlines()
+    fenced = _closed_fence_line_numbers(lines)
     prev = ""  # previous non-fence line (stripped); a setext title candidate
-    for line in text.splitlines():
+    for index, line in enumerate(lines):
+        if index in fenced:
+            prev = ""
+            continue
         s = line.strip()
-        if s.startswith("```") or s.startswith("~~~"):
-            in_fence = not in_fence
-            prev = ""
-            continue
-        if in_fence:
-            prev = ""
-            continue
         # Setext/RST underline: "=" (level 1) or "-" (level 2) directly under a
         # title line at least as long as the underline.
         if (
--- a/tests/test_unbalanced_code_fence.py
+++ b/tests/test_unbalanced_code_fence.py
@@ -128,6 +128,54 @@ class TestFenceCharacterMustMatch:
         assert _closed_fence_line_numbers(lines) == {0, 1, 2}
 
 
+class TestAcceptedOverCountCost:
+    """Pins the cost of the trade this fix makes, so it changes deliberately.
+
+    Leaving an unterminated fence's contents in the scan means code lines are
+    read as prose, and some of them can look like headings. The ATX branch
+    rejects digit-led and all-punctuation titles, but the **setext** branch has
+    no equivalent guard: any line sitting above a run of `-` or `=` becomes a
+    heading. Underlined section titles are common in `--help` output, which is
+    exactly what lives inside code fences in technical Markdown.
+
+    The over-count is accepted on purpose. An extra section is visible and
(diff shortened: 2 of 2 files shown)


virgiliojr94:
This is the right fix and the right trade — malformed input should degrade, not truncate, same as the unterminated RTF destination groups in #109. Checked out locally: 365 passed, `ruff check .` clean, and the truncation reproduces on `master`.

I measured the cost side of the trade myself, because "at worst adds a spurious heading candidate" deserved a number. A 4-section handbook whose third section holds an unclosed fence containing ordinary CLI output:

````
## Beta
```
$ tool --help
Options
-------
more code

## Gamma
...
````

```
master:   2 sections (Gamma and Delta swallowed)
this PR:  5 sections (all 4 real ones, plus "Options")
```

The spurious one is `Options` promoted by the `-------` under it. Worth being precise about why: the digit-led and all-punctuation guards you cite live on the **ATX** branch only — the setext branch has no equivalent, so any line above a run of `-` or `=` inside a now-unfenced block becomes a heading. Underlined section titles are extremely common in `--help` output, which is exactly what sits inside code fences in the technical Markdown this fallback runs on.

I still prefer 4→5 over 4→2 — an over-count is visible and survivable, a truncation is neither. But I want the cost pinned rather than discovered later, so: add that case as a test asserting the current number, with a comment saying the over-count is the accepted trade. If it ever changes, it should change deliberately.

Two answers to your notes:

- **RED evidence:** the behavioural before/after is enough. No need to split the direct-unit tests into a follow-up.
- **CommonMark fence length:** leave it out, as you proposed. It only matters for nested fences and it is not what this PR is about.

Add the test and I'll merge.

**Comment by Hotragn** [comment]
Test added in `b08f18f`, using your case verbatim.

Your diagnosis is exactly right, and I checked it rather than taking it on trust — the promotion is on the setext branch, and it is `Options` above the `-------`. Numbers against both trees:

```
master (unfixed)      2      Gamma and Delta swallowed
this branch           5      4 real sections + "Options"
fence closed          4      on both — no over-count when well-formed
```

I had not appreciated that the digit-led and all-punctuation guards are ATX-only. That asymmetry is the whole reason the cost exists, so I put it in the docstring rather than just recording the number — the next person to look at this should not have to re-derive why an unfenced block can mint a heading when `## 5 Setup` cannot.

Three assertions rather than one:

- `test_over_counts_by_one_setext_promotion` — the exact current number, so a change is deliberate.
- `test_every_real_section_survives` — `>= 4`, which is the point of the trade and would fail on `master`.
- `test_same_document_with_the_fence_closed_is_exact` — 4, showing the over-count is confined to malformed input.

Both your other calls taken as-is: no follow-up splitting the direct-unit tests, and no CommonMark fence-length check.

`366 passed, 3 skipped`, `ruff check .` clean.

One thing your analysis surfaces that is out of scope here but worth writing down: the setext branch having no guard is a standing weakness independent of fences. `Options` / `-------` in ordinary prose — a plain-text book with underlined headings, or `--help` output pasted outside a fence — mints a heading today, on `master`, with no fence involved. If you want, I can look at giving setext the same reject-list the ATX branch has (all-punctuation, digit-led) as a separate PR. Happy to leave it alone too; I would rather ask than widen this one.

PR #136, opened 2026-08-11
fix(extract): count supplementary-plane CJK in the token estimate

Hotragn:
> **Updated after review:** the range originally stopped at U+3134F, the end of the **Extension G** block, so Extension H (U+31350-U+323AF) still fell through. Widened to Planes 2-3 end to end; probes and labels corrected. See the review thread.

## Summary

`_CJK_RE` only covered the Basic Multilingual Plane, so Unified Ideographs Extension B and later (U+20000+) fell through to the whitespace-word branch — the exact ~1000× undercount #103 set out to fix, one plane up. 2,000 characters of Ext B/C/D/E estimated at **1 token** instead of 1,333.

## Type of change

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does

- **Fixes:** supplementary-plane ideographs were not recognised as CJK, so a space-less run of them counted as a single "word" in the cost pre-flight.
- **Improves:** a SIP-only text goes from **1** estimated token to the correct **1,333** per 2,000 characters; a realistic 90/10 BMP/SIP mix goes from ~10% low to exact.

## Motivation & context

Closes n/a — found while auditing the #103 estimate; no existing issue. This is a direct completion of that fix rather than a new direction.

The characters are not exotic in practice. Extensions B–I carry classical Chinese, Cantonese, Hong Kong and Taiwan place and personal names, and Japanese 人名用漢字 (the name-use kanji set). A book drawing on them is under-costed in proportion to how many it uses.

`estimate_tokens` feeds the Step 5 cost pre-flight, which is how a user decides whether a book is affordable to convert. Under-reporting is the harmful direction — it invites a run that then costs far more than quoted.

## How it works

One range added to the character class: `\U00020000-\U0003134F`, the ideographic supplementary planes (Planes 2-3).

It deliberately **stops short of the rest of the plane.** Emoji (U+1F600), mathematical alphanumerics (U+1D400) and regional indicators (U+1F1E6) are astral but are not ideographs — pulling them in would mis-scale ordinary Latin text that happens to contain them. `TestNonCjkSupplementaryPlanesExcluded` pins that boundary, and `test_latin_with_a_single_sip_character` checks that one rare ideograph in Latin prose does not flip the estimate onto the wrong scale.

Rejected alternative: `unicodedata.category(ch) == "Lo"`. It would catch the extensions, but also every other non-cased letter — Arabic, Hebrew, Thai, Devanagari — all of which *are* whitespace-delimited and are correctly handled by the word branch today.

## Evidence

**Before / after**, equal-length texts:

```
                        chars    before        after
BMP ideographs only      2000    1333 tokens   1333 tokens
Ext B/C/D/E only         2000       1 token    1333 tokens
90% BMP + 10% SIP        2000    1201 tokens   1333 tokens
```

**Tests** — new `tests/test_cjk_supplementary_plane.py`, 22 cases in three groups:

- `TestSupplementaryPlaneCounted` — SIP matches BMP for equal length, a SIP-only text is not one token, the estimate tracks `CJK_CHARS_PER_TOKEN`, mixed-plane text is consistent, and seven parametrized extension-range boundary probes (Ext B start/end, C, D, E, G, I).
- `TestExistingBehaviourPreserved` — BMP unchanged, short CJK/Korean/Japanese samples, Latin unaffected, empty text, and Latin-plus-one-ideograph.
- `TestNonCjkSupplementaryPlanesExcluded` — emoji, math alphanumerics and regional indicators still take the word branch.

RED against unmodified `master`: **11 failed, 11 passed** — a proper assertion-based RED (no new symbol imported). GREEN with the fix:

```
$ python -m pytest tests/ -q
369 passed, 3 skipped in 4.58s

$ ruff check .
All checks passed!
```

## Screenshots / output

The before/after table above is the output; this is a numeric estimate with nothing visual to capture.

## Checklist

- [x] `ruff check .` is clean

## Notes for reviewers

**On the upper bound.** I used U+3134F (end of Extension G) rather than the end of the plane. If you would rather not have to revisit this when a future Unicode release adds Extension J, `\U00020000-\U0003FFFF` covers the whole ideographic supplementary area and still excludes emoji and math alphanumerics, which live in Plane 1. That is a one-character change and I am happy either way — I went with the precise range so the comment stays truthful about what it covers.

**Not covered on purpose:** CJK Radicals Supplement (U+2E80–U+2EFF) and Kangxi Radicals (U+2F00–U+2FDF) are BMP but outside the current class. They appear in dictionaries and typographic discussion rather than running prose, so I left them alone rather than widen the scope of a fix aimed at the plane boundary. Easy follow-up if you want them.

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -53,8 +53,12 @@ from book_to_skill.sanitize import sanitize_extracted_text
 # fullwidth forms. These are not whitespace-delimited, so counting "words" on a
 # Chinese/Japanese book collapses it to a handful of tokens; count them directly.
 #
-# The last range is the supplementary plane, U+20000-U+3134F — Unified
-# Ideographs Extension B through I. Classical Chinese, Cantonese, Hong Kong and
+# The last range is Planes 2 and 3 (U+20000-U+3FFFF), the ideographic
+# supplementary planes, taken end to end rather than enumerated block by block
+# so a future extension does not silently fall through the way Extension H
+# (U+31350-U+323AF) did. Nothing non-ideographic lives up here: emoji,
+# mathematical alphanumerics and regional indicators are all in Plane 1, which
+# this range does not touch. Classical Chinese, Cantonese, Hong Kong and
 # Taiwan place/personal names, and Japanese 人名用漢字 all draw on it. Without it
 # those characters fell through to the whitespace-word branch, where a
 # space-less run of them counts as a single "word": the same ~1000x undercount
@@ -62,7 +66,7 @@ from book_to_skill.sanitize import sanitize_extracted_text
 _CJK_RE = re.compile(
     r"[　-〿぀-ヿ㐀-䶿一-鿿"
     r"가-힣豈-﫿＀-￯"
-    r"\U00020000-\U0003134F]"
+    r"\U00020000-\U0003FFFF]"
 )
 
 
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -52,9 +52,17 @@ from book_to_skill.sanitize import sanitize_extracted_text
 # CJK codepoints: ideographs + extensions, kana, hangul, CJK punctuation, and
 # fullwidth forms. These are not whitespace-delimited, so counting "words" on a
 # Chinese/Japanese book collapses it to a handful of tokens; count them directly.
+#
+# The last range is the supplementary plane, U+20000-U+3134F — Unified
+# Ideographs Extension B through I. Classical Chinese, Cantonese, Hong Kong and
+# Taiwan place/personal names, and Japanese 人名用漢字 all draw on it. Without it
+# those characters fell through to the whitespace-word branch, where a
+# space-less run of them counts as a single "word": the same ~1000x undercount
+# #103 fixed for the BMP, one plane up.
 _CJK_RE = re.compile(
     r"[　-〿぀-ヿ㐀-䶿一-鿿"
-    r"가-힣豈-﫿＀-￯]"
+    r"가-힣豈-﫿＀-￯"
+    r"\U00020000-\U0003134F]"
 )
 
 
--- a/tests/test_cjk_supplementary_plane.py
+++ b/tests/test_cjk_supplementary_plane.py
@@ -63,8 +63,14 @@ class TestSupplementaryPlaneCounted:
             (0x2A700, "Ext C start"),
             (0x2B740, "Ext D start"),
             (0x2CEB0, "Ext E start"),
+            (0x2EBF0, "Ext I start"),
             (0x30000, "Ext G start"),
-            (0x3134A, "Ext I range"),
+            (0x3134E, "Ext G end"),
+            # Extension H sits ABOVE Extension G, so a range that stopped at the
+            # end of G let 4,192 assigned ideographs fall through. Both ends are
+            # probed so the plane boundary cannot regress to a block boundary.
+            (0x31350, "Ext H start"),
+            (0x323AF, "Ext H end"),
         ],
     )
     def test_extension_ranges_are_covered(self, codepoint, name):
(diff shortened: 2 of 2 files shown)


virgiliojr94:
Good find, and the framing is right: this is the same undercount #103 fixed, one plane up. Checked out locally — 371 passed, `ruff check .` clean, and the before/after numbers reproduce.

The range has a hole, though, and it is the one your own "Notes for reviewers" anticipated.

`\U00020000-\U0003134F` stops at the end of the **Extension G** block (U+30000–U+3134F). **Extension H** (U+31350–U+323AF, 4,192 ideographs, assigned in Unicode 15.0 — present in the `unicodedata` shipped with the Pythons in our CI matrix) is above it and still falls through:

```
Ext B  U+20000   2000 chars -> 1333 tokens
Ext I  U+2EBF0   2000 chars -> 1333 tokens
Ext G  U+30000   2000 chars -> 1333 tokens
Ext H  U+31350   2000 chars ->    1 token   <- still the ~1000x undercount
Ext H  U+323AF   2000 chars ->    1 token
```

So "Extension B through I" in the code comment and the PR body is not what the range does, and the parametrized `(0x3134A, "Ext I range")` case is labelled wrong — U+3134A is Ext G. A book of rare ideographs that happens to draw on Ext H is still under-costed, which is the harmful direction you correctly identified.

Take the alternative you offered: `\U00020000-\U0003FFFF`. Planes 2 and 3 are the ideographic supplementary planes end to end — it covers Ext H and whatever Ext J lands as, and your exclusion argument is untouched, since emoji, math alphanumerics and regional indicators all live in Plane 1. Please also:

- fix the comment to say what it covers (Plane 2–3, the ideographic supplementary planes) instead of enumerating extensions that will drift,
- relabel the `0x3134A` case and add a real Ext H probe (`0x31350`),
- `TestNonCjkSupplementaryPlanesExcluded` stays as is — it is the test that makes the wider range safe, and it is why I am comfortable taking it.

The BMP radicals (U+2E80–U+2FDF) you left out: agreed, out of scope here.

With the range widened and the labels corrected, I'll merge.

Hotragn:
You are right, and I should have caught this — my own "Notes for reviewers" named the alternative and I still shipped the narrower range. Extension H sits **above** Extension G, not below, so stopping at the end of G left 4,192 assigned ideographs falling through. Fixed in `6ec4dab`.

Verified before and after, and the boundary is even sharper than the block numbers suggest — U+3134F is *unassigned*, while U+31350 is a named `CJK UNIFIED IDEOGRAPH`:

```
                        before   after
U+20000  Ext B           1333     1333
U+2EBF0  Ext I           1333     1333
U+30000  Ext G           1333     1333
U+31350  Ext H start        1     1333
U+323AF  Ext H end          1     1333
U+3FFFF  Plane 3 end        1     1333
```

Plane 1 still excluded, unchanged:

```
U+1F600 emoji               1 token
U+1D400 math bold A         1 token
U+1F1E6 regional indicator  1 token
```

All four points done:

- Range widened to `\U00020000-\U0003FFFF`.
- Comment now describes Planes 2–3 as the ideographic supplementary planes, and says explicitly that it is taken end to end *so a future extension does not silently fall through the way Extension H did* — rather than enumerating blocks that drift. It also states why Plane 1 is untouched, so the exclusion argument lives next to the range instead of only in a test.
- `0x3134A` relabelled (it was Extension G, not I), a real `Ext I start` probe added at `0x2EBF0`, and Extension H probed at **both** ends — `0x31350` and `0x323AF` — so the boundary cannot regress from a plane boundary back to a block boundary without a test failing.
- `TestNonCjkSupplementaryPlanesExcluded` untouched.

`372 passed, 3 skipped`, `ruff check .` clean.

Also corrected the PR body above, which carried the same wrong "Extension B through I" claim as the code comment.

PR #138, opened 2026-08-11
fix(discovery_tax): reuse the extractor's multilingual ToC detection

dex0shubham:
## Summary (Required)

`discovery_tax.py` reuses the extractor's `_chapter_number` "so discovery_tax and
the pipeline always agree" — but it duplicated ToC detection with a stale local
regex (`sumário|table of contents|contents|índice`, EN/ES/PT only). For a
French/German/Italian/Dutch/CJK book it missed the ToC and fell back to counting
the whole front matter, inflating the modelled discovery-loop token cost. This
reuses the extractor's `_TOC_PATTERN` instead.

## Type of change (Required)
- [x] ♻️ Refactor (no behavior change)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Fixes:** `tools/discovery_tax.py` now imports `_TOC_PATTERN` from
  `book_to_skill.utils` (alongside the `_chapter_number` it already reuses) and
  deletes the duplicated `TOC_RE`. `extract_toc` therefore detects the same
  multilingual + CJK ToC headers the extractor does, so the discovery-loop figure
  is measured from the real ToC on non-English books instead of the entire front
  matter.

The discovery-loop numbers back the README's headline token-savings claim
("measure, don't assert"). The stale regex silently over-counted the ToC for
every non-EN/ES/PT book. The file's own comment already states the principle:
reuse the extractor's detection so the two agree — this closes the gap for ToC.
Closes #n/a (no tracking issue).

`_TOC_PATTERN` is the same line-anchored (`^\s*X\s*$`, `IGNORECASE|MULTILINE`)
compiled regex `detect_structure` uses for `has_toc`; `extract_toc` calls
`.search()` on the front matter and slices from the match. No behavior change for
English books (`table of contents`/`contents` are in both). `re` is still used
elsewhere in the module.

## Evidence (Required)
- **Tests:** `test_extract_toc_detects_non_english_toc` — a German
  `Inhaltsverzeichnis` front matter now slices from the ToC heading and is shorter
  than the whole front matter (the old fallback returned everything).

```
$ pytest -q
348 passed, 3 skipped
$ ruff check .
All checks passed!
```

- [x] `ruff check .` is clean

`CHANGELOG.md` is left untouched — it's git-cliff-generated from the conventional
title now (#104), which carries this entry. Pure reuse of `_TOC_PATTERN`; no open
PR touches `discovery_tax.py`.

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -158,6 +158,7 @@ _KO_CHAPTER = re.compile(
 # line (^\s*X\s*$) so an inline "the contents of this chapter" never matches.
 _TOC_HEADERS = (
     "table of contents", "contents", "índice", "sumário",   # EN / ES / PT
+    "sumario",                                              # PT (no accent — OCR / accent-stripped, like indice below)
     "table des matières",                                   # French
     "inhaltsverzeichnis",                                   # German
     "indice", "sommario",                                   # Italian (no accent — distinct from índice above)
--- a/tests/test_book_to_skill.py
+++ b/tests/test_book_to_skill.py
@@ -712,6 +712,10 @@ class TestDetectStructure:
     def test_toc_spanish_accented(self):
         assert detect_structure("Índice\n1 Introducción")["has_toc"] is True
 
+    def test_toc_portuguese_unaccented(self):
+        # OCR / accent-stripped Brazilian PDFs leave "Sumario" without the accent.
+        assert detect_structure("Sumario\n1 Introdução")["has_toc"] is True
+
     def test_toc_traditional_chinese(self):
         assert detect_structure("目錄\n第一章")["has_toc"] is True
 
--- a/tests/test_discovery_tax.py
+++ b/tests/test_discovery_tax.py
@@ -120,3 +120,12 @@ class TestDiscoveryTaxOrdering:
         assert skill < d_best < dump, (skill, d_best, dump)
         assert d_best <= d_loop, (d_best, d_loop)
         assert skill < d_loop
+
+
+def test_extract_toc_detects_non_english_toc():
+    # German ToC: the stale local regex missed it and returned the whole front
+    # matter; reusing the extractor's _TOC_PATTERN slices from the ToC heading.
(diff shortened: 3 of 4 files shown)


virgiliojr94:
Thanks — reusing `_TOC_PATTERN` is the right call, and it's the same argument the file already makes for `_chapter_number`: one detector, one answer. Checked out locally, `pytest -q` is green (350 passed) and `ruff check .` is clean.

One thing to fix before I merge, because the swap is not purely additive.

The old local `TOC_RE` matched `sum[áa]rio` — **both** accented and unaccented. `_TOC_HEADERS` carries only `"sumário"`. So this PR gains FR/DE/IT/NL/CJK and loses unaccented Portuguese:

```
front matter        master   this PR
"Sumário"           sliced   sliced
"Sumario"           sliced   NOT sliced   <- regression
"Inhaltsverzeichnis"  no     sliced
"目 录"                no     sliced
```

Unaccented `Sumario` is not hypothetical — it is what an accent-stripping extraction pass or a poorly OCR'd Brazilian PDF leaves behind, and it is the language of a good share of this project's users.

The fix belongs upstream rather than back in `discovery_tax.py`: add `"sumario"` to `_TOC_HEADERS` in `book_to_skill/utils.py` (next to `"indice"`, which is already there for exactly this "same word, different accent" reason) with a short comment, plus a case in the extractor's ToC tests. The pipeline then gains the coverage too, which is the point of the shared detector — `discovery_tax` and the extractor should not disagree in either direction.

With that, this is a merge.

dex0shubham:
Good catch — fixed upstream as you suggested: added `"sumario"` to `_TOC_HEADERS` in `book_to_skill/utils.py` (with a comment noting the accent-stripped/OCR case, mirroring `indice`/`índice`), plus a `test_toc_portuguese_unaccented` extractor test. So both the pipeline and `discovery_tax` gain unaccented Portuguese — one detector, one answer, no regression. `pytest -q` green (349 passed), `ruff check .` clean. Thanks @virgiliojr94!

PR #139, opened 2026-08-11
feat: improve Persian chapter detection

HamedFarazi:
## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## What it does

- Improves Persian chapter detection for word-based ordinals from 1–34.
- Supports `هجدهم` and the common `هیجدهم` spelling variant.
- Supports compound ordinals such as `بیست و یکم` and `سی و چهارم`.
- Supports Persian, Arabic-Indic, and ASCII digits.
- Supports PDF-extracted glued chapter titles where the space between the chapter ordinal and title is lost.
- Adds regression and integration tests for Persian chapter detection.

## Motivation & context

Persian books can use word-based chapter numbers instead of digits, including compound ordinals.
PDF text extraction can also alter spacing or spelling, causing existing chapter detection to miss valid chapters.

This change improves detection for Persian books while preserving the existing behavior for other languages.

## Evidence

- The test suite detects all 34 Persian chapters in the target book.
- `pytest`: 357 passed, 4 skipped
- `ruff check .`: All checks passed

## Checklist

- [x] Tests added/updated
- [x] Existing tests pass
- [x] Lint checks pass
- [x] No unrelated changes
- [x] Documentation/PR description updated

## Summary

- Improve Persian chapter detection for word-based ordinals from 1–34
- Support `هجدهم` and common `هیجدهم` spelling variant
- Support compound ordinals such as `بیست و یکم` and `سی و چهارم`
- Support Persian, Arabic-Indic, and ASCII digits
- Support PDF-extracted glued chapter titles
- Add regression and integration tests

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -154,24 +154,73 @@ _KO_CHAPTER = re.compile(
     r"^\s*(?:#{1,6}\s+)?제\s*([0-9]+)\s*[장편절관](?:\s*의\s*[0-9]+)?(?:\s*$|[.:\-]|\s+\S)"
 )
 
-# Persian chapter headings: "فصل ۱", "فصل اول", "بخش ۲: مفاهیم", "فصل ۱۰".
-# Labels are فصل (chapter) and بخش (section/part). Digits may be ASCII, Persian
-# (U+06F0–U+06F9), or Arabic-Indic (U+0660–U+0669); int() parses all three.
-# Word numerals cover the common 1–10 ordinals only (no "بیست و یکم" composition).
-# Markdown "#" prefixes are handled by `_chapter_number`'s second pass (Issue #91),
-# not duplicated here. Trailing guard mirrors Korean: EOL, punctuation, or a
-# spaced title — Persian has no letter case for a Latin-style `_HEADING_TAIL`.
-_FA_WORD_NUMERALS = {
-    "اول": 1, "دوم": 2, "سوم": 3, "چهارم": 4, "پنجم": 5,
-    "ششم": 6, "هفتم": 7, "هشتم": 8, "نهم": 9, "دهم": 10,
-}
+# Persian chapter headings: "فصل ۱", "فصل اول", "بخش ۲: مفاهیم",
+# "فصل بیست و یکم", "فصل اولجایی…" (PDF glue). Labels are فصل / بخش.
+# Digits may be ASCII, Persian (U+06F0–U+06F9), or Arabic-Indic (U+0660–U+0669);
+# int() parses all three. Word numerals use a small ordinal map (1–34) with
+# longest-prefix matching so compounds ("بیست و یکم") and teens ("یازدهم") stay
+# maintainable — no giant alternation regex. Markdown "#" prefixes are handled
+# by `_chapter_number`'s second pass (Issue #91). Digit tails still require EOL /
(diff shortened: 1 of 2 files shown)


virgiliojr94:
Thanks for this — Persian word ordinals are a real gap, and the implementation is better shaped than the problem usually gets: a small ordinal map with longest-prefix matching instead of a giant alternation, the compounds built rather than enumerated, and the trailing guard modelled on the Korean one for the same reason (no letter case to lean on). 360 passed, ruff clean.

I checked the false-positive surface before the merge, and the glued form has one. Persian words that *begin* with a short ordinal get read as chapter headings:

```
فصل اولویت‌ها   ("chapter priorities")     -> 1
فصل اولیه       ("preliminary chapter")    -> 1
فصل دومینو      ("chapter domino")         -> 2
فصل سومین       ("the third chapter")      -> 3
```

All four are ordinary phrases, and `فصل اولیه` in particular is the kind of thing a technical book has as a real section title. Each becomes a spurious chapter, and since `_chapter_number` feeds `chapters_detected` and the structure map, the run reports a book that does not exist.

The cause is narrow: every one of them comes from a **short** ordinal — `اول`, `دوم`, `سوم`. A long ordinal is not a plausible prefix of another word; `یازدهم` or `بیست و یکم` followed immediately by more letters is extraction damage, not vocabulary.

So: keep the glue for teens and compounds, and require a separator (space, punctuation, ZWNJ, or end of line) after the ones 1–10. That keeps the PDF case you actually hit — which I assume was a long-form heading, but tell me if it was `فصل اول` glued, because then this needs a different answer — and drops all four cases above.

Two smaller things while you are in there:

- Add the false positives as tests, asserting `None`. They are the cases most likely to regress, and the next person to widen the glue should be told why it is narrow.
- `سی ام` and `سی‌ام` are both mapped for 30, but `بیست ام`/`بیست‌ام` are not, while `بیستم` is. If that asymmetry is deliberate (i.e. nobody writes it that way), a one-line comment saves the next reader the trip.

The rest I would take as is: rejecting `فصل الشتاء` and `در فصل قبل گفتیم` is exactly right, and English and Portuguese detection is untouched — I verified both.

HamedFarazi:
Thanks! Addressed the false-positive cases for short Persian ordinals (1–10) and added regression tests. Long-form/compound glued ordinals remain supported.

virgiliojr94 (approved):
Verified the revision. All four false positives now return `None`, and the cases that should still work do:

```
فصل اولویت‌ها  -> None      فصل اول            -> 1
فصل اولیه      -> None      فصل یازدهم         -> 11
فصل دومینو     -> None      فصل بیست و یکم     -> 21
فصل سومین      -> None      فصل هجدهمیک جاسوس  -> 18
```

The separator requirement is applied where it belongs — only to the 1–10 forms — so long-form and compound glue survives, which was the PDF case you were actually fixing. The regression tests assert `None` on all four and carry the reason in the docstring, so the next person to widen the glue is told why it is narrow. `بیست ام`/`بیست‌ام` now has the one-line comment explaining the asymmetry against `سی ام`.

Against `master` the new assertions fail (`فصل هجدهم عنوان` → `None`), so they cover the change and not something already true. 361 passed, 1 skipped; `ruff check .` clean.

✅ Approved.

PR #141, opened 2026-08-12
fix(security): harden extraction work directory against symlink and ownership attacks

johnnyc20:
## Summary (Required)
Replaces the bare `OUTPUT_DIR.mkdir(parents=True, exist_ok=True)` with `prepare_output_dir()`, which rejects a pre-planted symlink, a non-directory, or (on POSIX) a directory owned by another user at the predictable shared-`/tmp` extraction path, and tightens the directory to mode `0o700`.

## Type of change (Required)
- [x] 🔒 Security

## What it does (Required)
- **Fixes:** the extraction work directory was created with `mkdir(exist_ok=True)` against a predictable path under `/tmp`. Any local user could pre-plant a symlink, a non-directory, or a directory they own there ahead of the run, letting them read or tamper with extracted document text via a shared-tmp race/ownership attack.
- **Improves:** `prepare_output_dir()` now refuses all three cases with a clear `ExtractionError`, and creates/tightens the directory to mode `0o700` so it isn't group/world readable even if it does need reuse.

Closes n/a — found while reviewing the extraction pipeline's use of a fixed, predictable output path in shared temp space.

`prepare_output_dir(path)`: if `path` exists, `os.path.islink()` and `os.path.isdir()` gate against symlinks and non-directories; on POSIX, `os.stat().st_uid` is compared to `os.getuid()` to reject a directory owned by another user. On the accepted path it calls `os.chmod(path, 0o700)` (creating the directory first via `mkdir(parents=True)` if it doesn't exist yet), so both a fresh directory and a pre-existing directory the caller owns end up at the same restrictive mode.

## Evidence (Required)
- **Tests:** `tests/test_output_dir_security.py` (new) — asserts a fresh directory is created at mode `0o700`; a symlink at the target path raises `ExtractionError` matching "symbolic link"; a non-directory raises matching "not a directory"; a pre-existing world-writable (`0o777`) directory the caller owns gets tightened back to `0o700`; and (POSIX-only) a directory owned by a different uid raises matching "owned by a different user".
- **Before / after:** before, `OUTPUT_DIR.mkdir(parents=True, exist_ok=True)` would silently follow a pre-planted symlink or reuse a directory with any permissions/ownership. After, `prepare_output_dir()` raises `ExtractionError` on the symlink/non-directory/wrong-owner cases and the directory it hands back is always mode `0o700`.

- [x] `ruff check .` is clean
- [x] `python3 tools/validate_skill.py SKILL.md` passes (if `SKILL.md` changed) — n/a, `SKILL.md` not touched

This branch originally also patched `sanitize.py` to strip U+2060 (word joiner) from extracted text. That's dropped from this PR — `master`'s invisible-codepoint sweep (since this branch was first opened) already covers U+2060 and considerably more, so it was superseded upstream.

CodeQL flags `tests/test_output_dir_security.py:40` (`os.chmod(target, 0o777)`) as "overly permissive file permissions." That line intentionally simulates a pre-existing world-writable directory, inside pytest's isolated `tmp_path`, specifically to assert `prepare_output_dir()` tightens it back to `0o700` — i.e. it's exercising the fix, not introducing the vulnerability. Flagging for a maintainer to dismiss the alert or point me at this repo's preferred suppression convention, since I didn't see an existing one to follow.

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -679,6 +679,32 @@ def extract_single_file(input_path: Path, extraction_mode: str, install_mode: st
     }
 
 
+def prepare_output_dir(path: Path) -> None:
+    """Create the work directory, guarding against two shared-tmp risks:
+    a pre-planted symlink at a predictable path, and reusing a directory
+    another user already owns (either could expose or tamper with the
+    extracted document text, which may be sensitive).
+    """
+    if path.is_symlink():
+        raise ExtractionError(
+            f"Refusing to use {path}: it is a symbolic link, not a real "
+            "directory. Remove it or set BOOK_SKILL_WORKDIR to a private path."
+        )
+    if path.exists():
+        if not path.is_dir():
+            raise ExtractionError(f"Refusing to use {path}: it exists and is not a directory.")
+        if hasattr(os, "getuid"):
+            owner_uid = path.stat().st_uid
+            if owner_uid != os.getuid():
+                raise ExtractionError(
+                    f"Refusing to use {path}: it is owned by a different user "
+                    f"(uid {owner_uid}). Set BOOK_SKILL_WORKDIR to a private directory."
+                )
+            os.chmod(path, 0o700)
+    else:
+        path.mkdir(parents=True, mode=0o700)
+
+
 def print_banner() -> None:
     """Print the attribution banner. Done here (not only in SKILL.md) so it
     shows on every run regardless of how the agent invokes extraction."""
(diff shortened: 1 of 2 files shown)


johnnyc20:
CodeQL flagged `tests/test_output_dir_security.py:40` (`os.chmod(target, 0o777)`) as a high-severity "overly permissive file permissions" alert.

That's a false positive in context: the line simulates a pre-existing world-writable directory inside pytest's isolated `tmp_path` fixture, specifically so the very next assertion can verify `prepare_output_dir()` tightens it back to `0o700`. It's exercising the fix, not introducing the vulnerability.

I didn't see an existing inline-suppression convention in this repo, so I'm flagging rather than guessing one. Happy to add whatever suppression format you prefer, or this can just be dismissed as a false positive in the code scanning UI.

PR #142, opened 2026-08-12
fix(extractor): use trafilatura for real boilerplate detection in HTML

albatrossflyon-coder:
## Summary (Required)
`extract_html_content()` only strips `<script>`/`<style>`/`<head>` on both the bs4 and stdlib-fallback paths — neither does real main-content vs. page-chrome detection, so repeated site nav/footer/ad content pollutes every extraction. Adds `trafilatura` as the new primary extraction path (kept optional), falling through to the existing bs4/stdlib chain on any failure.

## Type of change (Required)

## What it does (Required)
- **Fixes:** `extract_html_content()` extracting repeated page chrome (nav/footer/ads) alongside real article content, since neither the bs4 path nor the stdlib fallback do main-content detection — they only strip `script`/`style`/`head`.
- **Improves:** boilerplate duplication in extracted text, measured against a real 61-page corpus (see Evidence).

Closes n/a — found while running the converter against 61 real pages scraped from personalmba.com (Josh Kaufman's *The Personal MBA*, official free-excerpt site) across two chapters, as part of evaluating book-to-skill for a real conversion.

`trafilatura.extract()` is purpose-built for "find the article, discard the chrome" — neither `bs4.BeautifulSoup.get_text()` nor the stdlib `_HTMLTextExtractor` attempt this, they only strip specific tags. `trafilatura` is tried first; on a missing dependency, a `None`/whitespace-only result (a page it can't confidently extract from), or any parse exception, it falls through to the existing bs4 path and then the stdlib parser — same graceful-degradation chain the function already used. Kept as an optional dependency (`pip install book-to-skill[html]`) so the base install stays dependency-free, matching how `beautifulsoup4` is already handled under the `epub` extra.

## Evidence (Required)
- **Tests:** `tests/test_html_boilerplate_extraction.py` (new) — 5 tests covering the fallback contract deterministically (trafilatura available/unavailable/returns-None/returns-whitespace/raises). trafilatura's own boilerplate-detection heuristics need realistic page-scale content to exercise meaningfully and would be flaky against a tiny synthetic fixture, so that behavior is verified against the real 61-page corpus below instead of a unit test.
- **Before / after:** ran `extract_html_content()` against the same 61 real personalmba.com pages (Chapters 1–2, "Value Creation" + "Marketing") before and after this change, counting occurrences of the site's known repeated footer text (ad block + author bio + copyright/trademark notice):

```
OLD (current bs4 path):  183 boilerplate hits across 61 pages
NEW (trafilatura path):   69 boilerplate hits across 61 pages

Breakdown of the 69 remaining:
  - ad block + copyright/trademark notice: 0 hits (fully eliminated)
  - author bio: dropped from 61 -> 4 hits
  - site masthead/tagline ("Master the Art of Business"): 65 hits --
    one line, once per page, sitting structurally where a real
    article lead-in/subtitle would (not nav/footer chrome), so
    trafilatura reasonably keeps it. Tried favor_precision=True too;
    barely moved (65), not worth the tradeoff.
```

Full test suite: 415 passed, 5 skipped (pre-existing, unrelated to this change), 3 failed — all 3 in `tests/test_output_dir_security.py`, added by the most recent commit on `master` (#141) before this branch was cut; confirmed via `git log` that this branch touches zero lines of that file. They assert POSIX-style `st_mode` permission bits and fail on Windows/NTFS, unrelated to HTML extraction.

**Update:** the initial push had the 4 trafilatura-mocking tests unguarded, so they hard-failed with `ModuleNotFoundError` on every CI Python version (CI installs only `pytest`, no extras — trafilatura is optional, by design). Fixed by guarding each with `pytest.importorskip("trafilatura")`, matching the existing bs4 skip idiom already used in `test_html_block_boundaries.py`. Verified both ways: with trafilatura installed, all 5 tests run and pass; without it (matching the real CI tier), 4 skip and 1 (the sys.modules-mocked "not installed" case) still runs and passes.

- [x] `pytest -q` is green on a clean checkout (excluding the 3 pre-existing unrelated failures above)
- [x] `ruff check .` is clean
- [x] `python3 tools/validate_skill.py SKILL.md` passes (if `SKILL.md` changed) — n/a, `SKILL.md` not touched

The remaining 69 boilerplate hits (all one masthead line) are a known, honest limitation, not an oversight — see the Evidence breakdown above. Happy to look at a more targeted fix for that specific line if it's worth the added complexity, but wanted to keep this PR to the real, measured win rather than chase a marginal gain.


virgiliojr94:
Thanks — and first, separately from this PR: you found a regression I shipped this morning. Those three failures in `tests/test_output_dir_security.py` were real, they hit every Windows contributor since #141 merged, and our CI is `ubuntu-latest` only so nothing caught them. Fixed and merged in #155. Reporting a failure in a file your branch never touched, instead of ignoring it as noise, is the kind of thing that keeps a project honest — thank you.

On this PR. The problem is real: neither `bs4.get_text()` nor the stdlib extractor attempts main-content detection, and for HTML sources that means every page's nav and footer land in the corpus. Your measurement is the right kind — 61 real pages, a named corpus, counted occurrences, and an honest accounting of what remains (the masthead that trafilatura reasonably keeps, and that `favor_precision` did not move). 418 passed on the branch.

Two things before I can take it.

**1. Seventeen packages.**

```
$ uv pip install --dry-run trafilatura
Resolved 17 packages
 + babel, certifi, charset-normalizer, courlan, dateparser, htmldate,
   justext, lxml, lxml-html-clean, python-dateutil, pytz, regex, six,
   tld, trafilatura, tzlocal, urllib3
```

lxml, a date parser, a timezone database and a URL classifier, to strip page chrome. That is not an argument against the library — it is excellent at what it does — but it is an argument for stating the cost where the user decides. The extra is the right call and you made it; what is missing is that `docs/install.md` lists the optional extras and this one arrives without a word about its weight. One line under the `[html]` extra saying it pulls a full HTML-processing stack is enough for someone on a constrained machine to choose the bs4 path deliberately.

**2. The new primary path is invisible.**

It is not registered in `DEPENDENCY_GROUPS`, so `--check` still reports HTML as `bs4 → fallback available` and never mentions that a better extractor exists or whether it is installed. And the run itself says nothing: every other parser in this project announces its choice —

```
Trying python-docx... OK
Trying stdlib DOCX parser... OK
  [warn] extract_with_pdftotext failed: …
```

— while this one silently returns one of two materially different documents. That matters more here than elsewhere, because the difference between the trafilatura result and the bs4 result is not cosmetic: it is whether the corpus contains the site's footer on every page. A user comparing two runs would have no way to know which one they got.

We merged #150 today for exactly this class of silence in chapter detection, so I would rather not add a new instance of it in the same week. What I want: register the module in the HTML group so `--check` reports it, and print which extractor answered, in the existing style.

Neither is large, and the measurement you already have is the hard part — it is done and it is convincing.

One question rather than a request: `include_tables=True` is right for this project, but `include_formatting=False` drops emphasis and inline code markers. For a book excerpt that is probably correct; for API documentation converted from HTML it might not be. Did the 61-page corpus have anything code-shaped in it, or is that untested territory?

albatrossflyon-coder:
Answering the open question first: **untested territory.** The 61-page corpus is exclusively personalmba.com prose (two book chapters) — no code blocks, inline code, or API-doc-style content anywhere in it. Confirmed by checking the corpus directly, not going from memory. `include_formatting=False`'s effect on code-shaped HTML genuinely hasn't been exercised by this measurement, so your instinct was right.

Both required changes are done:

1. **`docs/install.md`** now has a line under the pip install example noting the `[html]` extra pulls trafilatura's full 17-package HTML-processing stack (lxml, date parser, timezone database, URL classifier).
2. **`DEPENDENCY_GROUPS`** now lists trafilatura as the primary HTML module (bs4 second) — `--check` reports it correctly now. `extract_html_content()` also prints which extractor actually answered, mirroring `extract_docx()`'s "Trying X..." style, so a run no longer silently returns one of two materially different documents.

Also rebased onto current master to pick up #155 — full suite is green except one `test_output_dir_security.py` case that needs symlink-creation privileges Windows doesn't grant without Developer Mode/elevation; unrelated to this PR's diff and moot on your `ubuntu-latest` CI.

virgiliojr94:
Both required changes landed, and the answer to the formatting question is the right one — "untested territory, confirmed by checking the corpus rather than from memory" is a better answer than a guess. `--check` now reports the module and the run says which extractor answered. That part is done.

Two things left, both mechanical.

**1. `BUILDLOG.md` doesn't belong in this repo.**

The diff adds a 66-line file that is your own working log — local Windows paths (`C:\Repos\External\book-to-skill`), notes about which of your clones is which, references to an internal vault (`[[feedback-check-pr-template-before-contributing]]`), and a running account of your contribution sessions including other PRs. That is a useful document; it is just yours, not the project's. Merged here it becomes a file every future contributor sees and nobody can maintain, and it makes this PR two changes instead of one.

Please drop it from the branch. Everything in it that upstream needs is already in the PR description.

**2. Stray marker in a code comment.**

`parsers/html.py` opens the new block with `# ponytail: prints mirror extract_docx()'s ...`. That prefix means nothing in this codebase. The sentence after it is worth keeping — just remove the label.

With those two out, this is ready. The measurement was the hard part and it was done properly.

PR #146, opened 2026-08-13
style: run prettier to format style files

marcosvile:
## Summary (Required)

This PR formats the affected style files with Prettier and adds `.venv` to `.gitignore` to prevent local Python virtual environments from being tracked.

## Type of change (Required)

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [x] 🧹 Chore / CI / tooling

## What it does (Required)

- **Fixes:** Prevents local `.venv` directories from being included in Git tracking.
- **Improves:** Applies consistent Prettier formatting to the affected style files.
- **Adds:** `.venv` to `.gitignore` for local Python virtual environments.

Keeps repository formatting consistent and prevents local Python virtual environment files from being accidentally committed.

Closes #n/a — no issue was created for this maintenance change.

## How it works (Required)

Ran Prettier against the affected style files and added `.venv` to `.gitignore`. No application logic or runtime behavior was changed.

## Evidence (Required)

- **Tests:** No tests were added or changed because this PR contains no functional changes.
- **Before / after:** Prettier completed successfully and `.venv` is now ignored by Git.

```text
Prettier: completed successfully
.gitignore: .venv added
```

- [x] `ruff check .` is clean


virgiliojr94:
Thanks for taking the time on this — and I'm going to decline it, with the reasoning laid out so it doesn't read as arbitrary.

**The `.venv` half is already handled.** `uv` writes a `.gitignore` containing `*` inside the virtualenv it creates, so `.venv/` self-ignores — it has never appeared in `git status` here and has never been committed:

```
$ cat .venv/.gitignore
*
$ git log --all -- .venv     # (empty)
```

Adding it to the repo's `.gitignore` is harmless but solves nothing.

**The formatting half is the real question, and my answer is no for now.** Not because the output is wrong — I checked, and Prettier did nothing destructive here: the 11 `og:`/`twitter:` meta tags in `overrides/main.html` all survived and the Jinja inside their attributes is intact. The reasons are about timing and adoption:

1. **This project does not use Prettier.** No config, no CI step, no hook. A one-time reformat that nothing enforces is churn with a half-life: the next three PRs merge un-Prettiered content and the repo drifts back, having spent 437 lines of diff to get there.

2. **`SKILL.md` grows by 46 lines** (647 → 693, mostly blank lines after headings). It is already 200 lines over the 500-line guideline the project's own validator warns about, and I have publicly committed in #102 to cutting that file down rather than up. Taking a change that grows it, in the name of style, is the wrong direction on the one file where length is a standing debt.

3. **Four open PRs touch the same files** — #125, #126, #134 and #147 all edit `README.md`, `SKILL.md` or `docs/`. A 15-file reformat means every one of them rebases through formatting churn to land content changes. That cost falls on other contributors, not on us.

**If you want to pursue it, here is the path I would take it through:** open an issue proposing the formatter — which tool, which config, which files in scope, and enforced by a CI job so it stays true. Formatting decisions stick only when a machine applies them. Once that is agreed, the reformat lands in one commit at a moment when the queue is empty, so it conflicts with nothing. I would look at that seriously; markdown here is hand-formatted and inconsistent in places, and you are right about that.

What I would not want either way: Prettier over `SKILL.md`. That file is read by an agent step by step, and its line breaks and table shapes are load-bearing in ways a general-purpose formatter has no way to know.

Closing this one. The `.gitignore` line is welcome as a standalone one-liner if you would like to send it — I'd take that on its own, even though it is a no-op today, because someone using `python -m venv` instead of `uv` would not get the self-ignoring file.

PR #162, opened 2026-08-14
fix(extractor): treat bracketed paths as literals

### LINKED ISSUE #158: Paths containing square brackets are silently unreadable (glob branch has no literal fallback)
## Summary

`resolve_input_paths()` routes any path containing `[` to the glob branch, even when that path exists literally on disk. `glob.glob()` interprets `[...]` as a character class, finds nothing, and the file is dropped — with a misleading "No supported files found" error.

This makes the tool unable to read **any** file whose name contains square brackets. That is not an edge case for one common workflow: Zotero (with the ZotMoov add-on) names every attachment `Author - Title [year].pdf`, so an entire personal library is unreadable.

## Reproduce

```bash
$ ls -l 'Landon - Building Great Sentences [2013].pdf'
-rw-r--r--@ 1 user staff 1.4M  'Landon - Building Great Sentences [2013].pdf'

$ python3 scripts/extract.py 'Landon - Building Great Sentences [2013].pdf' --mode text
ERROR: No supported files found matching: Landon - Building Great Sentences [2013].pdf

# same file via a bracket-free symlink
$ ln -s 'Landon - Building Great Sentences [2013].pdf' landon.pdf
$ python3 scripts/extract.py landon.pdf --mode text
   Sources : 1 processed
   Total Pages: 254
   Words   : 78,831
```

`[2013]` is a valid glob character class matching one of `2`, `0`, `1`, `3` — so it never matches the literal directory entry.

## Cause

`book_to_skill/utils.py:537`

```python
if any(char in path_str for char in ("*", "?", "[")):
    glob_matches = glob.glob(path_str, recursive=True)
```

The wildcard check is purely lexical. It has no fallback when the glob expands to nothing but the literal path exists.

## Suggested fix

Try the literal path first, and only treat the argument as a pattern if it does not exist:

```python
p = Path(path_str)
if not p.exists() and any(char in path_str for char in ("*", "?", "[")):
    glob_matches = glob.glob(path_str, recursive=True)
    ...
```

This keeps glob support intact (a real pattern does not exist as a literal path) and fixes literal paths containing `[`, `]`, `?` or `*`. `glob.escape()` is an alternative, but the exists-first check also covers `?` and `*` in filenames, which are legal on macOS and Linux.

A regression test would be a file literally named `book [2013].pdf` passed as a single argument.

## Impact

Silent and total for affected filenames: the error message points at the input rather than at the path handling, so the natural conclusion is that the file is corrupt or unsupported. It took a symlink experiment to find the real cause.

Environment: macOS 15.6, Python 3.11, book-to-skill at `master`.

NgoQuocViet2001:
## Summary

Treat an existing input path as a literal before checking it for glob metacharacters. This lets files such as `book [2013].pdf` resolve normally without changing actual glob expansion.

## Type of change

## What it does

- **Fixes:** existing filenames containing square brackets being interpreted as glob patterns and resolving to no files.

## Motivation & context

Closes #158.

## How it works

The resolver enters the glob branch only when the literal path does not exist. Existing files continue through the explicit file path branch, while real patterns keep the same expansion, filtering, sorting, and deduplication behavior.

## Evidence

- **Tests:** added a regression test for a literal `book [2013].pdf`; the focused path/glob/tilde group passes 17 tests.
- **Before / after:** the new test returned `[]` before the fix and the resolved file path after it.

```
before: 1 failed (result: [])
after:  1 passed (result: [book [2013].pdf])
full:   470 passed, 7 skipped
ruff:   All checks passed
```

The only additional local result is the existing Windows symlink-creation test, which needs the OS symlink privilege and fails with WinError 1314 before reaching project code.

## Checklist

- [x] `ruff check .` is clean

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -534,7 +534,9 @@ def resolve_input_files(paths: list[str]) -> list[Path]:
         # the file/directory branch below see a real path.
         path_str = os.path.expanduser(raw_path)
         # Check if it has glob wildcards
-        if any(char in path_str for char in ("*", "?", "[")):
+        if not Path(path_str).exists() and any(
+            char in path_str for char in ("*", "?", "[")
+        ):
             glob_matches = glob.glob(path_str, recursive=True)
             # Sort expanded glob results deterministically
             expanded = []
--- a/tests/test_book_to_skill.py
+++ b/tests/test_book_to_skill.py
@@ -1206,6 +1206,13 @@ class TestDocxExtraction:
 class TestResolveInputFiles:
     """Additional edge-case tests for resolve_input_files."""
 
+    def test_existing_file_with_glob_metacharacters_is_literal(self, tmp_path):
+        target = _make_text_file(tmp_path / "book [2013].pdf")
+
+        result = resolve_input_files([str(target)])
+
+        assert result == [target.resolve()]
+
     def test_nonexistent_file_kept_for_error_reporting(self, tmp_path):
         """A nonexistent explicit path is kept so extract_single_file can report it."""
         fake = tmp_path / "nonexistent.pdf"


virgiliojr94 (approved):
✅ **Approved** — this is the cleanest fix for the bracket-path issue and arrived first (Aug 14). The one-line change (`if not Path(path_str).exists() and any(...)`) ensures literal paths containing glob metacharacters resolve correctly while preserving glob expansion for genuine patterns.

**Cluster context:** three PRs (#162, #168, #171) independently fixed the same issue. This one wins on:
- **Simplicity**: 1-line change vs larger refactors
- **Timing**: opened Aug 14, before #168 (Aug 15) and #171 (Aug 16)
- **Correctness**: regression test covers the exact case from #158

I'll merge this shortly and close the duplicates with credit to their authors. Thanks for the quick, focused fix!

cc @TheodoreAI @albatrossflyon-coder

PR #164, opened 2026-08-15
test(security): skip symlink assertion when unsupported

Stamina9:
## Summary (Required)

Keep the output-directory security test portable on hosts that cannot create directory symlinks. The security assertion still runs everywhere directory symlinks are available.

## Type of change (Required)

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [x] 🧹 Chore / CI / tooling

## What it does (Required)

- **Fixes:** `test_prepare_output_dir_rejects_symlink` failed during setup on Windows installations without Developer Mode or symlink privileges, before `prepare_output_dir()` was exercised.
- **Improves:** the Windows suite result from 1 failed / 470 passed / 6 skipped to 470 passed / 7 skipped on such hosts.
- **Adds:** an explicit directory-symlink capability check while preserving the security assertion on capable hosts.

The permission-bit assertions in this test module were made POSIX-only in #155, but the remaining symlink test still assumes every host may create a symlink. On Windows that operation commonly raises `WinError 1314`, making an otherwise green local suite fail for an environmental limitation.

Closes: n/a — discovered while verifying current `master` on Windows; no matching open issue or PR was found.

The test requests a directory symlink explicitly with `target_is_directory=True`. If the platform does not implement symlinks or the host refuses their creation, the test reports a capability-based skip. When creation succeeds, the existing assertion that `prepare_output_dir()` raises `ExtractionError` remains unchanged.

## Evidence (Required)

- **Tests:** `tests/test_output_dir_security.py` still asserts that a successfully created symlink is rejected; the full suite was rerun from a clean detached worktree.
- **Before / after:** reproduced on Windows 11 with Python 3.12.

```text
Before: 1 failed, 470 passed, 6 skipped
FAILED tests/test_output_dir_security.py::test_prepare_output_dir_rejects_symlink
OSError: [WinError 1314] A required privilege is not held by the client

After: 470 passed, 7 skipped
SKIPPED tests/test_output_dir_security.py: directory symlinks are unavailable on this host: [WinError 1314]

python -m ruff check .
All checks passed!

python tools/validate_skill.py SKILL.md
no Claude Code-breaking issues (1 pre-existing line-count warning)
```

- [x] `ruff check .` is clean

Suggested PR title: `test(security): skip symlink assertion when unsupported`

This changes test setup only; production symlink rejection behavior is untouched.

### DIFF
--- a/tests/test_output_dir_security.py
+++ b/tests/test_output_dir_security.py
@@ -30,7 +30,10 @@ def test_prepare_output_dir_rejects_symlink(tmp_path):
     real_dir = tmp_path / "real"
     real_dir.mkdir()
     link = tmp_path / "work"
-    link.symlink_to(real_dir)
+    try:
+        link.symlink_to(real_dir, target_is_directory=True)
+    except (NotImplementedError, OSError) as exc:
+        pytest.skip(f"directory symlinks are unavailable on this host: {exc}")
 
     with pytest.raises(ExtractionError, match="symbolic link"):
         prepare_output_dir(link)


virgiliojr94:
Thanks @Stamina9 — this is the right fix, merging it.

Wrapping `symlink_to` in `try/except` → `pytest.skip` keeps `test_prepare_output_dir_rejects_symlink` running wherever directory symlinks can actually be created, and only skips when the host genuinely can't make one — so we keep coverage on capable platforms instead of dropping the test wholesale.

Context: this overlaps #173, which touches the same file (`tests/test_output_dir_security.py`) for the same Windows-portability reason. Yours is the canonical fix for the symlink test; the two chmod/permission tests (`0o700`, tighten-existing) are separately POSIX-only, and I've asked #173 to cover just those. Credit to @dex0shubham for flagging them in parallel.

PR #165, opened 2026-08-15
docs: add Hermes Agent as a supported host

dineshr93:
## Summary (Required)

Adds Hermes Agent (`$HERMES_HOME/skills`, default `~/.hermes/skills`) as a supported host alongside Copilot CLI, Amp, Claude Code, and Codex. Docs-only change to `SKILL.md` (+16/−12): the converter's script lookup, skill-location ranking, and generated-skill destination table now recognize Hermes so a conversion run inside it works end-to-end instead of guessing at another host's root.

## Type of change (Required)

- [x] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)

- **Adds:** Hermes Agent as a first-class host in `SKILL.md`: frontmatter description + cross-agent notes, Skill Locations list (ranked #1), Step 2 `extract.py` lookup (`${HERMES_HOME:-$HOME/.hermes}/skills/book-to-skill/scripts/extract.py`, env-var aware so non-default profiles resolve too), and a Hermes row in the Step 5 `SKILLS_HOME` table + host-detection prompt.
- **Improves:** running the converter inside Hermes now writes generated book skills to the running host's skill root (`~/.hermes/skills/<name>/`) instead of probing only Copilot/Amp/Claude roots and potentially asking the user or picking a root their agent can't discover.

Closes #n/a — no issue filed; Hermes Agent is an Agent Skills-compatible host that was missing from the host lists, so its users had to hand-edit `SKILL.md` after installing (this PR's own install required exactly those edits).

No new steps added — only existing lists/tables extended with one host row each. The lookup path uses `${HERMES_HOME:-$HOME/.hermes}` so both default and profile-based installs resolve correctly, matching how Hermes resolves its home directory.

## Evidence (Required)

- **Tests:** no code files touched; full suite still green on the changed tree (`pytest -q` → 477 passed).
- **Before / after:** local preflight of the exact commands CI runs:

```
WARN  body: 705 lines > 500 (soft guideline for optimal performance)   # pre-existing, unchanged by this PR

$ pytest -q
477 passed in 0.45s

# ruff: docs-only diff, no code files changed
```

Also verified in practice on a Hermes install (symlinked checkout into `~/.hermes/skills/`): skill loads, and an end-to-end markdown extraction run completed successfully with the Step 2 lookup resolving to the Hermes path.

- [x] Tests added/updated for behavior changes *(none needed — docs-only, no code behavior change)*
- [x] `ruff check .` is clean
- [x] No raw book text shipped; no net `SKILL.md` bloat without justification *(+4 lines net, all host-list entries required for correct routing in Hermes)*

The Hermes row is placed first in the Skill Locations list and lookup loop because this PR targets a Hermes install; other hosts fall through to their existing paths unchanged. Happy to reorder or gate it differently if you'd rather keep alphabetical/host-precedence conventions.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -1,12 +1,13 @@
 ---
 name: book-to-skill
-description: "Converts books and documents (PDF, EPUB, DOCX, HTML, Markdown, plain text, RTF, MOBI/AZW with Calibre) into structured agent skills, extracting frameworks, mental models, principles, techniques, and anti-patterns. Use when the user wants to study a document through GitHub Copilot CLI, Amp, or Claude Code, apply an author's frameworks while working, or build a reusable knowledge base from a file."
+description: "Converts books and documents (PDF, EPUB, DOCX, HTML, Markdown, plain text, RTF, MOBI/AZW with Calibre) into structured agent skills, extracting frameworks, mental models, principles, techniques, and anti-patterns. Use when the user wants to study a document through Hermes Agent, GitHub Copilot CLI, Amp, or Claude Code, apply an author's frameworks while working, or build a reusable knowledge base from a file."
 ---
 
 <!--
 Cross-agent notes (informational; ignored by host agents):
-  - Compatible skill roots: GitHub Copilot CLI (~/.copilot/skills, ~/.agents/skills,
-    .github/skills, .claude/skills, .agents/skills), Amp (.agents/skills,
+  - Compatible skill roots: Hermes Agent ($HERMES_HOME/skills, default ~/.hermes/skills),
+    GitHub Copilot CLI (~/.copilot/skills, ~/.agents/skills, .github/skills,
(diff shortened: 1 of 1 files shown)


virgiliojr94:
Thanks for the PR! The integration makes sense technically, but Hermes has since added /learn, which now natively covers the book/large-document → knowledge-base skill workflow and explicitly builds on the pattern popularized by book-to-skill.

Because of that, adding Hermes-specific routing and paths here would give Hermes users two overlapping ways to do essentially the same thing and add host-specific maintenance on our side.

I think the better direction is to document Hermes as a native integration/adoption of the book-to-skill pattern rather than maintain a separate execution path inside Hermes.

https://x.com/Teknium/status/2085761587550519420

PR #166, opened 2026-08-15
fix(extractor): extract multi-column PDFs in reading order instead of -layout (#128)

### LINKED ISSUE #128: PDF: `pdftotext -layout` interleaves columns on multi-column documents, and the word-based token estimate hides the 26x size blowup

aryansk:
## What

Fixes #128. `pdftotext -layout` welds every physical line across all columns, so on a multi-column document consecutive sentences from different columns get interleaved — the issue's "sentences that never existed in the book" failure — while padding the output ~26x (96.8% whitespace on IRS Pub 17).

`extract_with_pdftotext()` now:

1. Runs `pdftotext -layout` first (unchanged for single-column docs).
2. Flags the result with the issue's own gutter heuristic — `looks_multicolumn()`: the fraction of substantive (40+ char) lines carrying a 4+ space internal run, sampled over the first 4000 lines, threshold 0.35.
3. When flagged, re-runs plain `pdftotext` (reading order) and returns that instead. Single-column books keep `-layout` exactly as before; both runs are cheap (seconds on 142 pages).

## Verified against the issue's repro (IRS Pub 17)

- Layout output: 1,641,750 chars, `looks_multicolumn` → True.
- Final extraction: 957,292 chars (~the issue's 957,757 no-flag baseline), **0 gutter lines** in the output, sentences contiguous in reading order (the standard-deduction passage reads correctly end to end).

## Tests

9 new tests in `tests/test_multicolumn_pdf_detection.py`: detector behavior (flagged, single-column, short-line, empty, threshold), and mode selection with mocked `pdftotext` (multi-column → 2 calls with reading-order second, single-column → 1 call keeping `-layout`, failed probe → None, missing binary → None without calling). Full suite: **485 passed, 1 skipped**; `ruff check` and `ruff format --check` clean.

I implemented the issue's suggested detection approach as proposed. I did not fold in the three "smaller changes" (char-based token estimates, padding collapse, multi-column warning) — happy to follow up on any of them separately if you want them.

### DIFF
--- a/book_to_skill/parsers/pdf.py
+++ b/book_to_skill/parsers/pdf.py
@@ -26,6 +26,30 @@ _ROMAN_1_99 = r"(?=[ivxl])(?:xc|xl|l?x{0,3})(?:ix|iv|v?i{0,3})"
 _PDF_PAGE_NUM = re.compile(rf"^\s*(?:\d{{1,4}}|{_ROMAN_1_99})\s*$", re.IGNORECASE)
 _PDF_HYPHEN_WRAP = re.compile(r"(\w)-\n(\w)")
 
+# A wide internal run of spaces — the column gutter. Multi-column `-layout`
+# output pads each physical line across all columns, so a substantive line
+# contains 4+ spaces between two non-space characters; single-column reading
+# order never does (#128).
+GUTTER = re.compile(r"\S[ ]{4,}\S")
+
+
+def looks_multicolumn(layout_text: str, threshold: float = 0.35) -> bool:
+    """True when `-layout` output looks like a multi-column page: under -layout
+    each physical line spans all columns, so a large fraction of substantive
+    lines carry a wide internal run of spaces (the gutter). Reading order has no
+    gutters, so the fraction is near zero on single-column documents.
+
+    Only substantive lines (40+ chars after stripping) are sampled, capped at
+    the first 4000 lines for a bounded cost on huge documents. Returns False on
+    empty or all-short input — a book with no long lines is not multi-column.
+    """
+    lines = [ln for ln in layout_text.splitlines() if len(ln.strip()) > 40]
+    if not lines:
+        return False
+    sample = lines[:4000]
+    gutter_lines = sum(1 for ln in sample if GUTTER.search(ln))
+    return gutter_lines / len(sample) >= threshold
+
 
 def clean_pdftotext(text: str) -> str:
     """Clean pdftotext '-layout' output (pages are form-feed delimited): drop
@@ -73,20 +97,48 @@ def clean_pdftotext(text: str) -> str:
     return _PDF_HYPHEN_WRAP.sub(r"\1\2", text)
 
 
-def extract_with_pdftotext(pdf_path: str) -> str | None:
+def _run_pdftotext(
+    pdf_path: str, extra_args: list[str], timeout: int
+) -> subprocess.CompletedProcess[str] | None:
+    """Best-effort single pdftotext invocation; returns None when pdftotext is
+    missing or the call fails outright."""
     if not shutil.which("pdftotext"):
         return None
     try:
-        pdf_path = os.path.abspath(pdf_path)
-        result = subprocess.run(
-            ["pdftotext", "-layout", pdf_path, "-"],
-            capture_output=True, text=True, timeout=120,
-            encoding="utf-8", errors="replace",
+        return subprocess.run(
+            ["pdftotext", *extra_args, os.path.abspath(pdf_path), "-"],
+            capture_output=True,
+            text=True,
+            timeout=timeout,
+            encoding="utf-8",
+            errors="replace",
         )
-        if result.returncode == 0 and result.stdout.strip():
-            return clean_pdftotext(result.stdout)
     except Exception as e:
-        print(f"  [warn] extract_with_pdftotext failed: {type(e).__name__}: {e}", file=sys.stderr)
+        print(f"  [warn] pdftotext failed: {type(e).__name__}: {e}", file=sys.stderr)
+        return None
+
+
+def extract_with_pdftotext(pdf_path: str) -> str | None:
+    """Extract text with pdftotext, choosing the mode per document (#128).
+
+    `-layout` keeps table columns aligned on single-column documents, but on a
+    multi-column page it welds every column's lines together: consecutive
+    sentences from different columns get interleaved, and the padding inflates
+    the output ~26x. Both extractions are cheap (seconds for a 140-page book),
+    so we run `-layout` first and re-run in reading order (no flag) only when
+    the gutter heuristic flags the document as multi-column.
+    """
+    layout = _run_pdftotext(pdf_path, ["-layout"], timeout=120)
+    if layout is not None and layout.returncode == 0 and layout.stdout.strip():
+        if looks_multicolumn(layout.stdout):
+            reading = _run_pdftotext(pdf_path, [], timeout=120)
+            if (
+                reading is not None
+                and reading.returncode == 0
+                and reading.stdout.strip()
+            ):
+                return clean_pdftotext(reading.stdout)
+        return clean_pdftotext(layout.stdout)
     return None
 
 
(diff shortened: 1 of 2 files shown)


virgiliojr94:
Thanks for tackling #128. The diff is focused and the reported IRS Pub. 17 result is useful.\n\nBefore I can arbitrate this cluster, please add a reproducible measurement path: identify the exact public fixture/version and command(s), then show the detector result plus `-layout` vs reading-order character/whitespace counts. The new regression test also needs to fail on `master` and pass on this head.\n\n#129 solves the same root cause, so I will select the measured implementation rather than merge both.

virgiliojr94 (commented):
**LGTM with one blocker** — solid fix for #128, clean implementation, but CI gate failing + 29 commits behind.

## What it does
Fixes multi-column PDF extraction (#128). `pdftotext -layout` welds columns together → interleaved sentences + 26× output bloat. Solution: probe with `-layout`, detect multi-column via gutter heuristic, re-run in reading order (plain `pdftotext`) when flagged.

## Implementation

### ✅ Correct
- **Gutter heuristic:** `GUTTER = r"\S[ ]{4,}\S"` detects 4+ space runs — present in `-layout` multi-column (padding between columns), absent in reading order
- **Threshold calibrated:** 0.35 (35% of substantive lines carry gutter) → empirically tested on IRS Pub 17 (issue repro)
- **Bounded cost:** samples first 4000 lines, only substantive (>40 chars) → O(1) on huge PDFs
- **Fallback preserved:** single-column keeps `-layout` (table alignment), multi-column gets reading order (sentence continuity)
- **Dual-run cheap:** both calls = seconds on 142-page book

### ✅ Evidence
- **Issue repro verified:** IRS Pub 17 → 1,641,750 chars (`-layout`) → flagged → 957,292 chars (reading order), 0 gutter lines, sentences contiguous ✓
- **Tests:** 9 new cases in `test_multicolumn_pdf_detection.py` — detector behavior (threshold, empty, short-line) + mode selection (mocked subprocess)
- **Empirical validation:** all detector edge cases pass ✓
- **CI:** 14/15 green, **1 FAIL: PR description check**

### ⚠️ Blocker: CI gate failing
```
PR description check (PR): FAILURE / fail
```

PR body missing required sections from template (likely "Type of change" checkboxes or "Checklist"). Fix PR description to match template before merge.

### ⚠️ Divergence: 29 commits behind master
Same as #170 — significant drift since 2026-08-15. **Rebase required before merge** to avoid conflicts.

## Code quality

### ✅ Good
- **Single responsibility:** `looks_multicolumn()` isolated, testable
- **No magic numbers:** threshold=0.35 documented, overridable
- **Graceful degradation:** missing pdftotext → None (no crash), failed probe → None
- **Backward compat:** single-column behavior unchanged

### Minor notes
- **`_run_pdftotext()` factoring:** clean extraction of subprocess call, reusable
- **Error messages updated:** "pdftotext failed" replaces "extract_with_pdftotext failed" (more specific)
- **Ruff formatting applied:** multi-line string formatting (cosmetic, good hygiene)

## Compliance
- ✅ One focused change (multi-column detection only)
- ✅ Tests added for behavior change
- ✅ Fixes real issue (#128 with repro)
- ✅ No SKILL.md growth
- ✅ No new runtime deps
- ✅ Backwards compat preserved (single-column path unchanged)
- ⚠️ **PR description incomplete** (CI gate red)
- ⚠️ **Rebase needed** (29 commits behind)

## Action required before merge
1. **Fix PR description** → add missing template sections (Type of change checkboxes, Checklist)
2. **Rebase on current master** → resolve 29-commit divergence
3. **Re-run CI** → verify green after rebase

Implementation is solid — only process gates blocking.

PR #167, opened 2026-08-15
fix(epub): report images that text-only extraction drops (#127)

aryansk:
## What

EPUB extraction is text-only, so every image in an EPUB was dropped without any signal. This PR implements the scope agreed in the issue thread: **count and report**, **warn when figure-borne content is likely**, and **say it in the generated skill**.

- **`count_epub_images()`** (`parsers/epub.py`): counts images from the OPF manifest (`media-type="image/..."` items), falling back to an archive extension scan when the manifest declares none. Degrades to 0 on error, exactly like `count_epub_chapters`.
- **Surfaced in three places**:
  1. extraction-time note to stderr (per source),
  2. the run report (`WARN: N image(s) in the source were not extracted — figures, diagrams and charts are unavailable.`),
  3. `metadata.json` as `"images_dropped": N` — per source row and consolidated at the top level (only when N > 0, so text-only sources are untouched).
- **SKILL.md**: the generated skill's Scope & Limits template now instructs the agent to state that figures in the source were not read when `metadata.json` reports `images_dropped`; the technical-mode message no longer promises Docling structure preservation for EPUB (Docling only ever runs on the PDF path).

## Why

A figure-heavy EPUB (226 images / 7.8 MB of the 8.4 MB in the issue's repro) produced a skill that was confidently incomplete — captions survived, their payload didn't, and nothing in the output said so. Per the maintainer's direction, book-to-skill is *not* going to read figures; this converts the silent loss into a stated limit without crossing that boundary.

## Validation

- 10 new tests (`tests/test_epub_images_reported.py`): manifest counting, extension fallback, zero for imageless EPUBs, unreadable-archive safety, `extract_single_file` plumbing (EPUB row, zero, non-EPUB absence), and `main()` metadata + run-report behavior (single and multi-source).
- Regression proven: with the count removed, 4/4 plumbing tests fail.
- Full suite: **486 passed, 1 skipped**; `ruff check` clean on all touched files.
- End-to-end smoke test with a 4-image EPUB: `[info] 4 image(s)...` at extraction, `WARN` line in the report, `images_dropped: 4` in metadata (top-level and per source).

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -117,7 +117,7 @@ Store the answer as `BOOK_TYPE`:
 - Option 3 → `BOOK_TYPE=text`
 
 **If `BOOK_TYPE=technical`**, inform the user before proceeding:
-> "📐 Technical mode selected — using Docling for structure-aware extraction (tables, code blocks, formulas preserved as markdown). This takes ~1.5s per page, so expect a few minutes for longer sources. Starting now…"
+> "📐 Technical mode selected — using Docling for structure-aware extraction of PDFs (tables, code blocks, formulas preserved as markdown). EPUB sources are extracted as text only: their images are never read, so figures are not preserved. This takes ~1.5s per page, so expect a few minutes for longer sources. Starting now…"
 
 **If `BOOK_TYPE=text`**, inform:
 > "📄 Text mode selected — using the fastest suitable extractor for each file type. Plain text/Markdown/HTML are usually ready in seconds; PDFs use pdftotext when available."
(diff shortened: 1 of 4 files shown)


virgiliojr94:
@aryansk Thank you for implementing EPUB image reporting! This PR is now a duplicate of #134 (opened Aug 10, five days before yours), which implements the same feature with a simpler extension-based count.

**What happened:**
- #134 (Aug 10): count by extension, warn when >5 images, 3 focused tests
- #167 (Aug 15, this PR): manifest-first with extension fallback, warn always, 10 tests

Both solve #127 correctly. I'm merging #134 because it arrived first, has all checks passing, and uses a simpler approach. Your PR's manifest-first counting is interesting — consider opening a follow-up "enhancement" PR for that specific improvement after #134 lands, or I can cherry-pick it if you prefer.

Closing this as a duplicate with full credit to you for independently implementing the fix. Thanks for your contribution!

cc @hydraxman (author of #134)

PR #170, opened 2026-08-16
feat(skill): split Steps 7-11 into a lazily-loaded GENERATION.md

albatrossflyon-coder:
## Summary (Required)
SKILL.md's own instructions weighed 8,891 tokens and loaded in full on every run, even when a run aborts before generation is ever reached. Split Steps 7-11, the Update/Fold-in Workflow, and Quality Rules into a lazily-loaded GENERATION.md, cutting the always-loaded cost to 4,249 tokens (52.2%).

## Type of change (Required)
- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ Feature (non-breaking change that adds capability)
- [x] ⚡ Performance (faster / cheaper, no behavior change)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling
- [ ] 💥 Breaking change (changes existing behavior or a public interface)

## What it does (Required)
- **Improves:** always-loaded token weight of SKILL.md, 8,891 → 4,249 tokens (52.2% cut, tiktoken `cl100k_base`, exact measurement), matching the issue's own "52%" framing.
- **Adds:** GENERATION.md, holding Steps 7-11 + Update/Fold-in Workflow + Quality Rules, loaded only once a run actually commits to generating (Step 2.5 confirmation for Modes 1-3, or Step 5's branch for Mode 4).
- **Fixes:** `tests/test_publish_visibility_gate.py`, which parsed SKILL.md by string search for the Step 11 publish/visibility safety gate — that section moved, so the test now reads GENERATION.md. The gate's prose is unchanged, only relocated.

## Motivation & context (Required)
Closes #169

## How it works (Required for anything non-trivial)
Moved Steps 7-11 + the Update/Fold-in Workflow + Quality Rules verbatim into GENERATION.md, replacing that content in SKILL.md with a short pointer.

The one non-obvious decision: Mode 4 (Update/Fold-in) skips Step 2.5's cost-estimate gate entirely — it runs Steps 0, 1, 1.5, 2, then jumps straight to Step 5 and into the Update/Fold-in Workflow. A naive "defer everything behind the Step 2.5 confirmation" design would have silently broken Mode 4, since it never passes through that gate. Both files call out both entry points explicitly (Step 2.5 confirmation for Modes 1-3; Step 5's branch for Mode 4), rather than relying on a single trigger point.

[...]

PR #171, opened 2026-08-16
fix(extractor): fall back to literal path when bracket-containing files exist

### LINKED ISSUE #158: Paths containing square brackets are silently unreadable (glob branch has no literal fallback)

albatrossflyon-coder:
## Summary (Required)
resolve_input_files() routed any path containing "[" to glob.glob() unconditionally, before checking whether it existed literally on disk. glob.glob() treats "[...]" as a character class, so a real filename like Zotero's default "Author - Title [year].pdf" naming matched nothing and was silently dropped. Fix checks literal existence first, falling back to glob expansion only when there's no literal match.

## Type of change (Required)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Fixes:** a literal file or directory whose name contains "[" (or "?"/"*") no longer gets silently dropped. Root cause: the glob-char check ran before any literal-existence check, so `glob.glob()` — which treats "[...]" as a character class — matched nothing and the path never reached the literal-path fallback that already existed for non-glob paths.

Closes #158

Restructured the branch condition from "has glob chars? / else" to "(has glob chars AND doesn't exist literally)? / is a directory? / else (literal file)". A literal path that exists — file or directory — now always takes that path, regardless of whether its name happens to contain glob metacharacters. Genuine glob patterns (no literal file/dir of that exact name) fall through to `glob.glob()` exactly as before.

Also removed a duplicate `Path(path_str)` construction (previously built once in the glob-miss branch and again in the else branch; now built once at the top of the loop) and renamed the inner glob-loop variable from `p` to `gp` to stop it shadowing the outer `p`.

## Evidence (Required)
- **Tests:** new `tests/test_bracket_path_literal_fallback.py`, 4 tests — a literal bracket-named file resolves, a literal bracket-named directory is walked, a genuine bracket-glob pattern (`ch[0-9].md`) still expands correctly, and a bracket path matching nothing either way still returns `[]` unchanged (confirms this fix doesn't touch that pre-existing, out-of-scope case).
- **Before / after:**

```
Before (on upstream/master):
>>> resolve_input_files(["Landon - Building Great Sentences [2013].pdf"])
[]   # file exists on disk, silently dropped

After (this branch):
>>> resolve_input_files(["Landon - Building Great Sentences [2013].pdf"])
[WindowsPath('.../Landon - Building Great Sentences [2013].pdf')]

pytest -q:
  Before this branch (clean upstream/master): 1 failed, 470 passed, 6 skipped
  After this branch:                          1 failed, 474 passed, 6 skipped
  (+4 = the new test file; same pre-existing unrelated failure both times:
   test_prepare_output_dir_rejects_symlink, Windows symlink privilege)

ruff check . -> All checks passed!
vuln-hunter scan_diff -> No findings.
```

- [x] `ruff check .` is clean
- [x] `python3 tools/validate_skill.py SKILL.md` passes (if `SKILL.md` changed) — n/a, this PR doesn't touch SKILL.md, so this requirement does not bind

No `BUILDLOG.md` entry, same as PR #170 — that's my own workflow's convention, not this project's.

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -533,32 +533,37 @@ def resolve_input_files(paths: list[str]) -> list[Path]:
         # Normalise "~" once, at the entry point, so both the glob branch and
         # the file/directory branch below see a real path.
         path_str = os.path.expanduser(raw_path)
-        # Check if it has glob wildcards
-        if any(char in path_str for char in ("*", "?", "[")):
+        p = Path(path_str)
+        has_glob_chars = any(char in path_str for char in ("*", "?", "["))
+
+        # A literal path that exists takes priority over glob expansion: "["
+        # is a valid character in a real filename (e.g. Zotero's default
+        # "Author - Title [year].pdf" naming), but glob.glob() treats it as
+        # opening a character class and silently matches nothing. Only
+        # expand as a glob pattern when there's no literal match.
+        if has_glob_chars and not p.exists():
             glob_matches = glob.glob(path_str, recursive=True)
             # Sort expanded glob results deterministically
             expanded = []
             for match in glob_matches:
-                p = Path(match)
-                if p.is_file() and p.suffix.lower() in SUPPORTED_EXTENSIONS:
-                    expanded.append(p.resolve())
+                gp = Path(match)
+                if gp.is_file() and gp.suffix.lower() in SUPPORTED_EXTENSIONS:
+                    expanded.append(gp.resolve())
             expanded.sort(key=lambda x: str(x).lower())
             resolved.extend(expanded)
+        elif p.is_dir():
+            # Sort expanded directory results deterministically
+            dir_files = []
+            for root, _, files in os.walk(p):
+                for file in files:
+                    file_path = Path(root) / file
+                    if file_path.suffix.lower() in SUPPORTED_EXTENSIONS:
+                        dir_files.append(file_path.resolve())
+            dir_files.sort(key=lambda x: str(x).lower())
+            resolved.extend(dir_files)
         else:
-            p = Path(path_str)
-            if p.is_dir():
-                # Sort expanded directory results deterministically
-                dir_files = []
-                for root, _, files in os.walk(p):
-                    for file in files:
-                        file_path = Path(root) / file
-                        if file_path.suffix.lower() in SUPPORTED_EXTENSIONS:
-                            dir_files.append(file_path.resolve())
-                dir_files.sort(key=lambda x: str(x).lower())
-                resolved.extend(dir_files)
-            else:
-                # Keep even if it doesn't exist so the error check can report it
-                resolved.append(p.resolve())
+            # Keep even if it doesn't exist so the error check can report it
+            resolved.append(p.resolve())
 
     # Deduplicate while preserving insertion order (user order for explicit files)
     seen = set()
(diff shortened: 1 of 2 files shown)


virgiliojr94:
@albatrossflyon-coder Thank you for the detailed fix and comprehensive test suite! This PR is now a duplicate of #162 (opened Aug 14), which implements a minimal 1-line fix for the same bracket-path issue.

**What happened:**
- #162 (Aug 14): `if not Path(path_str).exists() and any(...)` — smallest diff
- #168 (Aug 15): same logic, different order
- #171 (Aug 16, this PR): larger refactor + optimization (single `Path()` call, restructured branches)

All three solve the problem correctly. I'm merging #162 because it arrived first and changes the least code. Your PR's optimizations (avoiding duplicate `Path(path_str)` calls, clearer branch structure) are valuable — consider opening a follow-up "refactor" PR targeting those specifically after #162 lands, or I can cherry-pick them if you prefer.

Closing this as a duplicate with full credit to you for the thorough analysis and test coverage. Thanks for your contribution!

cc @NgoQuocViet2001 (author of #162) @TheodoreAI (author of #168)

PR #177, opened 2026-08-19
docs: fix the broken star history chart

FaintFlower:
The star history chart in the README is currently broken because the underlying stargazer API stopped working without a token. This switches the chart to a token-free provider so it renders again for all visitors. The change updates both the English and Russian READMEs.


virgiliojr94:
This was already planned since I have other ideas, but I've already removed it in #179; thanks for the contribution.

PR #178, opened 2026-08-19
fix(sanitize): strip the invisible carriers a Cf filter does not reach

NgoQuocViet2001:
## Summary (Required)

`sanitize.py` strips the Unicode tag block, but four neighbouring channels get through. Each one renders as nothing to the reviewer approving a generated skill while the model reads it in full — which is exactly the split the module exists to close.

## Type of change (Required)

- [ ] 📝 Docs only
- [x] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)

- **Fixes:** invisible text survives sanitisation through four channels the current predicate misses. Root cause: `is_invisible_codepoint` is organised around category `Cf` plus a few named letters, so it never reaches combining marks (`Mn`) or symbols (`So`) that also render as nothing.

A generated skill is approved by a human reading it. Any code point that draws nothing but is still delivered to the model breaks that review — the reviewer and the agent are reading different documents. The tag block was closed for this reason; these four carry the same payload capability.

Closes # n/a — found while reading `sanitize.py`, no existing issue.

`is_invisible_codepoint` is the only thing that changes. That is deliberate: extraction and `tools/scan_generated_skill.py` both call it, so they move together and the drift that `test_scanner_shares_the_extractor_predicate` guards against stays impossible.

The four channels, and why each belongs:

**Variation selectors** (U+FE00–U+FE0F, U+E0100–U+E01EF) — the tag-block trick moved to a block that survives more pipelines. Each selector after any base character carries one of 256 values, so a run of them encodes an arbitrary payload behind a single visible glyph. They are combining marks rather than format controls, which is why a `Cf`-shaped filter never saw them.

Dropping them costs only the emoji/text presentation hint on a character that already renders. That is a smaller loss than the module already accepts: U+200D is stripped today, which splits emoji ZWJ sequences outright.

**Interlinear annotation controls** (U+FFF9–U+FFFB) — a conforming renderer hides the annotation between the anchor and the terminator, so the reviewer and the agent read different text.

**U+2800 BRAILLE PATTERN BLANK** — a symbol, not a format control: it draws nothing and survives whitespace normalisation. U+2801 and the rest of the block have raised dots and are left alone.

**U+1D173–U+1D17A** — zero-width format characters that pad hidden text anywhere, not only inside musical notation.

Rejected alternative: matching on `unicodedata.category` alone. It would sweep in every `Mn` combining mark and destroy ordinary accented text.

## Evidence (Required)

- **Tests:** new cases in `tests/test_sanitize_extracted_text.py` cover a 16-selector payload reduced to its base character; that U+FDFF / U+FE10 / U+E00FF / U+E01F0 — the code points just outside each new range — are kept; that a dotted braille pattern is kept while U+2800 goes; an annotation-wrapped injection losing its controls; and that `scan_generated_skill.py` flags every newly covered code point.
- **Before / after:** measured on `master`, then on this branch.

```
                                                  master   this branch
U+FE0F  VARIATION SELECTOR-16          Mn        kept     stripped
U+E0100 VARIATION SELECTOR-17          Mn        kept     stripped
U+2800  BRAILLE PATTERN BLANK          So        kept     stripped
U+FFF9  INTERLINEAR ANNOTATION ANCHOR  Cf        kept     stripped
U+1D173 MUSICAL SYMBOL BEGIN BEAM      Cf        kept     stripped
U+E0000 (tag block)                              stripped stripped

Boundaries held (still kept, both before and after):
U+FDFF, U+FE10, U+E00FF, U+E01F0, U+2801 BRAILLE PATTERN DOTS-1

$ python -m pytest tests/ -q
491 passed, 12 skipped

$ python -m pytest tests/test_sanitize_bidi_controls.py tests/test_sanitize_extracted_text.py -q
39 passed

$ uvx ruff check . --output-format concise
All checks passed!
```

## Screenshots / output (Required if behavior or output changes)

The behaviour change is terminal-visible only and is the before/after table above — there is no UI surface for it.

- [x] `ruff check .` is clean
- [x] `python3 tools/validate_skill.py SKILL.md` passes (if `SKILL.md` changed) — n/a, `SKILL.md` is untouched
- [x] PR title follows Conventional Commits (`fix:`, `feat:`, `docs:`…)

Deliberately out of scope: I did not touch the `Cf`-based core of the predicate, only extended what it reaches. If you would rather see the four ranges expressed as a named table instead of inline conditions, say so and I will restructure — it seemed better to match the file's existing shape than to impose a new one in a fix.

Apologies for the original title — `sanitize:` is not a Conventional Commits type, so the title and description checks failed on my own metadata rather than on the code. Both are corrected here; the 12 code checks were green throughout.

### DIFF
--- a/book_to_skill/sanitize.py
+++ b/book_to_skill/sanitize.py
@@ -48,14 +48,13 @@ _BIDI_CONTROL_CODEPOINTS = frozenset({
 })
 
 # 3. Characters that are not format controls (so a category-based filter misses
-#    them) but still render as blank width. Unlike a space they are letters or
-#    symbols, so they survive whitespace normalisation and can pad hidden text.
+#    them) but still render as blank width. Unlike a space they are letters, so
+#    they survive whitespace normalisation and can pad hidden text.
 _INVISIBLE_LETTER_CODEPOINTS = frozenset({
     0x115F,  # HANGUL CHOSEONG FILLER
     0x1160,  # HANGUL JUNGSEONG FILLER
     0x3164,  # HANGUL FILLER
     0xFFA0,  # HALFWIDTH HANGUL FILLER
-    0x2800,  # BRAILLE PATTERN BLANK — a symbol that draws no dots
 })
 
 _INVISIBLE_CODEPOINTS = (
@@ -84,16 +83,20 @@ _VARIATION_SELECTOR_RANGES = (
     (0xE0100, 0xE01EF),  # VARIATION SELECTOR-17 .. -256 (supplement)
 )
 
-# 6. Interlinear annotation controls. A conforming renderer hides the annotation
+# 7. Interlinear annotation controls. A conforming renderer hides the annotation
 #    between the anchor and the terminator, so text a human never sees is still
 #    read in full by the model — the same split this module exists to close.
+# 6. Deprecated format controls. Category Cf and Default_Ignorable, with no
+#    legitimate use in extracted book prose; contributed by #182.
+_DEPRECATED_FORMAT_RANGE = (0x206A, 0x206F)
+
 _ANNOTATION_CODEPOINTS = frozenset({
     0xFFF9,  # INTERLINEAR ANNOTATION ANCHOR
     0xFFFA,  # INTERLINEAR ANNOTATION SEPARATOR
     0xFFFB,  # INTERLINEAR ANNOTATION TERMINATOR
 })
 
-# 7. Musical beaming and phrasing controls: zero-width format characters that
+# 8. Musical beaming and phrasing controls: zero-width format characters that
 #    can pad hidden text anywhere, not only in musical notation.
 _MUSICAL_FORMAT_RANGE = (0x1D173, 0x1D17A)
 
@@ -106,6 +109,8 @@ def is_invisible_codepoint(codepoint: int) -> bool:
     the scanner then warns about — or worse, neither layer covers it.
     """
     if codepoint in _INVISIBLE_CODEPOINTS or codepoint in _ANNOTATION_CODEPOINTS:
+        return True
+    if _DEPRECATED_FORMAT_RANGE[0] <= codepoint <= _DEPRECATED_FORMAT_RANGE[1]:
         return True
     if _TAG_BLOCK_START <= codepoint <= _TAG_BLOCK_END:
         return True
--- a/book_to_skill/sanitize.py
+++ b/book_to_skill/sanitize.py
@@ -48,13 +48,14 @@ _BIDI_CONTROL_CODEPOINTS = frozenset({
 })
 
 # 3. Characters that are not format controls (so a category-based filter misses
-#    them) but still render as blank width. Unlike a space they are letters, so
-#    they survive whitespace normalisation and can pad hidden text.
+#    them) but still render as blank width. Unlike a space they are letters or
+#    symbols, so they survive whitespace normalisation and can pad hidden text.
 _INVISIBLE_LETTER_CODEPOINTS = frozenset({
     0x115F,  # HANGUL CHOSEONG FILLER
     0x1160,  # HANGUL JUNGSEONG FILLER
     0x3164,  # HANGUL FILLER
     0xFFA0,  # HALFWIDTH HANGUL FILLER
+    0x2800,  # BRAILLE PATTERN BLANK — a symbol that draws no dots
 })
 
 _INVISIBLE_CODEPOINTS = (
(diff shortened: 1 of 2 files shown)


virgiliojr94:
Thanks for the focused hardening work. I re-ran both change sets against `master` and their respective heads.\n\nThese PRs overlap on `U+FFF9–U+FFFB`, but neither subsumes the other:\n- #178 adds variation-selector coverage (and `U+2800` handling) that #182 does not cover.\n- #182 adds `U+206A–U+206F` coverage that #178 does not cover.\n- The PRs also intentionally disagree on whether `U+2800` should be stripped.\n\nI am keeping both open while we settle the `U+2800` contract, then will consolidate the non-overlapping coverage with attribution to both contributors.

NgoQuocViet2001:
Thanks for running both and for saying plainly which parts don't overlap — that makes the consolidation easy to reason about.

On the `U+2800` contract, I want to concede something first, because I think @dex0shubham is more right than my PR gives them credit for.

**U+2800 is the braille space.** In real braille text it is the word separator: `⠓⠑⠇⠇⠕⠀⠺⠕⠗⠇⠙` is "hello world". Strip it and you get `⠓⠑⠇⠇⠕⠺⠕⠗⠇⠙` — one run-together word. My PR strips it unconditionally, so on a book with a braille table or an accessibility chapter it destroys content. That is a real regression and "it draws nothing" was too blunt a test on my part.

**But the carrier concern is also real.** `U+2800` is used for steganography precisely because it is blank-but-not-whitespace: it survives `.strip()` and whitespace normalisation, so a run of it pads hidden text through exactly the gap this module exists to close. Keeping it unconditionally leaves that open.

So I don't think either blanket rule is right, and I'd rather not have the consolidation pick a loser. **Concrete proposal for the contract:**

> `U+2800` is stripped only when it has no braille neighbour. If any character in the same run is `U+2801–U+28FF` (a pattern with actual raised dots), the `U+2800` is a space in real braille content and is kept. In isolation — surrounded by ASCII, or by nothing — it is a carrier and goes.

That gives @dex0shubham's case the answer they want (`⠓⠑⠇⠇⠕⠀⠺⠕⠗⠇⠙` is untouched, and their test asserting the braille blank is kept still passes as written if the fixture has dotted neighbours), and it closes the padding vector, which a plain keep does not. It also matches how the module already reasons — `U+FE00–U+FE0F` are judged by what they attach to, not in isolation.

I'm happy to implement it here so the consolidation has one settled behaviour to take, and equally happy for it to land in #182 instead if you'd rather the `Cf` groups and this one travel together — no attribution concerns from me either way.

If you'd prefer to keep this simple and just say "don't strip `U+2800`", say so and I'll drop it from this PR today. The variation-selector coverage (`U+FE00–U+FE0F`, `U+E0100–U+E01EF`) and `U+1D173–U+1D17A` are the parts of #178 that #182 doesn't have, and I'd rather those land unblocked than hold them hostage to one code point.

One correction to my own PR while I'm here: the description says the branch is rebased on latest `master`. That was true when I wrote it; it is worth re-checking before you consolidate, since #182 landed after. Happy to rebase on request.

virgiliojr94:
Decision: preserve `U+2800 BRAILLE PATTERN BLANK` unconditionally. It is semantic braille content, not a control or Default_Ignorable character; the concrete corruption risk outweighs the unproven padding concern. Please update this PR to remove `U+2800` from the stripped set, retain the variation selectors and `U+FFF9–U+FFFB`, and add the non-overlapping `U+206A–U+206F` coverage from #182. Add explicit regressions for an isolated `U+2800` and a braille sequence being preserved.\n\nThe neighbour heuristic is intentionally deferred: the sanitizer predicate is shared by extraction and the generated-skill scanner, so context-sensitive handling would widen this focused hardening change. If padding becomes evidenced, we can detect anomalous runs separately.\n\nOnce updated, I will validate the consolidated PR and close #182 with credit.

NgoQuocViet2001:
Updated and pushed. Your call on `U+2800` is the right one — corruption of real content is a certain cost, the padding vector was a hypothetical, and those do not trade evenly. And the reason for deferring the neighbour heuristic is better than my proposal: I was treating `is_invisible_codepoint` as if it only served extraction, when the scanner shares it, so context-sensitivity would have leaked scope into both.

**What changed**

- `U+2800` removed from the stripped set, preserved unconditionally.
- Neighbour heuristic dropped entirely, not implemented behind a flag.
- `U+206A–U+206F` added, the non-overlapping coverage from #182.

**Retained**, since all three are category `Cf` or `Default_Ignorable`, which is the line your decision draws: the variation selectors (`U+FE00–U+FE0F`, `U+E0100–U+E01EF`), the interlinear annotation controls (`U+FFF9–U+FFFB`), and the musical format controls (`U+1D173–U+1D17A`). Say the word if you would rather the musical range travel separately.

**New regressions**

- an isolated `U+2800` between ASCII survives with `removed == 0`
- a full braille word sequence (`⠛⠑⠇⠇⠕⠀⠺⠕⠗⠇⠙`, blank included) round-trips byte-for-byte
- `U+206A`, `U+206C`, `U+206F` are stripped
- the range's upper neighbour `U+2070` is kept, while `U+2069` stays stripped as the bidi control it already was

That last pair is worth flagging: I first wrote the boundary test against both neighbours and it failed, because `U+2069` is POP DIRECTIONAL ISOLATE and section 1 already strips it deliberately. The test now asserts that explicitly rather than papering over it.

**Verification**

```
python -m pytest tests/ -q          528 passed, 12 skipped
uvx ruff check .                    All checks passed!
```

Also rebased onto current `master` as promised, so the branch is current for the consolidation.

PR #180, opened 2026-08-21
fix(extractor): reject punctuation-only setext titles, matching the ATX branch

Hotragn:
> Follow-up to your review on #135: you traced the over-count there to the setext branch having no guard, and I asked whether to give it the ATX reject-list as a separate PR. This is that PR, scoped to the unambiguous half only.

## Summary

`_structural_chapter_count` has two heading branches. The ATX branch rejects a title with no word character — that is what stops a `=====` table border or a `***` thematic break being counted. The setext branch had no equivalent, so the **identical string** was rejected as `## ***` and accepted as `***` sitting above a row of `-`.

## Type of change

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does

- **Fixes:** a punctuation-only line above a setext underline was counted as a chapter heading. Two thematic breaks in a row (`***` over `---`) produced a heading titled `***`.
- **Improves:** the two branches now judge the same string the same way — asserted directly, not just implied.

## Motivation & context

Closes n/a — surfaced by your analysis on #135.

The docstring already promises this behaviour: *"thematic breaks, table borders, and front-matter `---` do not match"*. That was only true while the underline was **shorter** than the line above it, because `len(s) >= len(prev)` was doing the work. A long rule defeats it:

```
paragraph + SHORT "---" rule   -> 2   (correct)
paragraph + LONG  "-----" rule -> 3   (length guard defeated)
```

So this is a documented-versus-actual mismatch, not a new rule.

## How it works

One condition added to the setext branch, the same test the ATX branch already applies:

```python
and re.search(r"\w", prev)
```

## Evidence

Two real sections in every fixture, so any count above 2 is a phantom:

```
title       setext before   setext after   ATX (unchanged)
***               3              2          2
+-----+           3              2          2
.......           3              2          2
|||||             3              2          2
* * *             3              2          2
###               3              2          2
-- --             3              2          2
```

**Tests** — new `tests/test_setext_punctuation_guard.py`, 25 cases in four groups:

- `TestPunctuationOnlySetextTitleRejected` — the seven titles above (parametrized), each also asserted **equal to the ATX result** for the same string, plus two thematic breaks in a row and the long-underline case.
- `TestRealSetextHeadingsStillCounted` — word titles, titles carrying punctuation (`Chapter One -- Beginnings`), digit-and-dot titles (`1.2 Scope`), a CJK title (`\w` is Unicode-aware), and `snake_case_title`.
- `TestSharedUnderscoreGap` — see below.
- `TestUnaffectedBehaviour` — no-heading documents, the length guard, blank line above an underline.

RED against `master`: **16 failed, 9 passed**. GREEN:

```
$ python -m pytest tests/ -q
512 passed, 11 skipped in 5.21s

$ ruff check .
All checks passed!
```

## Screenshots / output

The before/after table above is the output; this is a count with nothing visual to capture.

## Checklist

- [x] `ruff check .` is clean

## Notes for reviewers

**Deliberately NOT fixed — the ambiguous half.** A title *with* word characters still counts: `Options` above `-------`, and a plain-text table header like `Name   Value` above `------------`. Those are indistinguishable from a real setext heading by shape alone, and you already accepted that cost on #135. I did not want to smuggle a heuristic in behind a consistency fix.

**One shared gap I found and left alone.** `___` is a valid thematic break, but `_` **is** a word character to `\w` — so `___` is counted by *both* branches, before and after this PR:

```
"___" as setext -> 3      "## ___" as ATX -> 3
```

I could have used `[^\W_]` instead and killed it, but that would make setext *stricter* than ATX and re-open the divergence this PR closes. Fixing it properly means changing both branches, which is a separate decision — so I pinned the parity in `TestSharedUnderscoreGap` instead, with a comment saying as much. Happy to send that follow-up if you want it.

**Unrelated, still open from earlier:** my #71 comment about `validate_docx_xml_safety` decoding every archive member five times is still unanswered. That PR looks stalled — say the word and I will send it as a standalone PR against `master` instead.

### DIFF
--- a/book_to_skill/utils.py
+++ b/book_to_skill/utils.py
@@ -388,6 +388,13 @@ def _structural_chapter_count(text: str) -> int:
             and prev
             and not _SETEXT_UNDERLINE.match(prev)
             and len(s) >= len(prev)
+            # A title made only of punctuation is never a chapter. Two thematic
+            # breaks in a row ("***" over "---"), an ASCII box rule, a row of
+            # dots, or a table border sitting above an underline all reach this
+            # point. The ATX branch below already rejects them with the same
+            # test; the setext branch had no equivalent, so the identical string
+            # counted as a heading here and not there.
+            and re.search(r"\w", prev)
         ):
             depth = 1 if s[0] == "=" else 2
             levels.setdefault(depth, set()).add(prev.lower())
--- a/tests/test_setext_punctuation_guard.py
+++ b/tests/test_setext_punctuation_guard.py
@@ -1,0 +1,131 @@
+r"""A setext title made only of punctuation is not a chapter heading.
+
+`_structural_chapter_count` has two heading branches. The ATX branch rejects a
+title with no word character — that is what keeps a `=====` table border or a
+`***` thematic break from being counted:
+
+    if title and re.search(r"\w", title):
+
+The setext branch had no equivalent. So the *identical string* was rejected as
+`## ***` and accepted as `***` sitting above a row of `-`. Two thematic breaks
+in a row, an ASCII box rule, a row of dots, or a punctuation table border above
+an underline all minted a phantom heading.
+
+The function's own docstring claims "thematic breaks, table borders, and
+front-matter `---` do not match", which was only true while the underline was
+shorter than the line above it.
+"""
+
+import sys
+from pathlib import Path
+
+import pytest
+
+ROOT_DIR = Path(__file__).resolve().parent.parent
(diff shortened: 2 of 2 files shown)


virgiliojr94 (approved):
Reproduced: 16 punctuation cases fail on `master`; all 25 focused tests and the full suite pass on this head. Focused fix, clean CI.

PR #182, opened 2026-08-21
fix(sanitize): strip deprecated and interlinear-annotation format controls

dex0shubham:
## Summary (Required)

The invisible-code-point scrub in `book_to_skill/sanitize.py` missed nine
`Cf` / Default_Ignorable format controls that render as nothing — the same
invisible-format-control attack shape it already blocks (group 1). They passed
straight through the extraction scrub, and since the generated-skill scanner
shares `is_invisible_codepoint`, went unflagged there too. This adds them.

## Type of change (Required)
- [ ] 📝 Docs only
- [x] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Adds** `U+206A–U+206F` (deprecated format controls: symmetric-swapping,
  Arabic-form-shaping, digit-shapes) and `U+FFF9–U+FFFB` (interlinear annotation
  anchor/separator/terminator) to the sanitizer's invisible-code-point blocklist,
  as a new commented group. Because `is_invisible_codepoint` is shared, this
  hardens **both** the extraction scrub and the generated-skill scanner in one
  change.

These are all category `Cf`, `Default_Ignorable`, and render as nothing, with no
legitimate use in extracted book prose — the exact "invisible spacer / format
control" vector the file's comments describe (hidden text a human reviewer can't
see but the model reads). The original list simply didn't include them; this is
the same "a code point was missed" gap as the CJK-plane fixes (#136), one layer up.
Conservative: I deliberately did **not** add visible `So` characters like braille
blank (U+2800) or musical symbols, which are real content.
Closes #n/a (no tracking issue).

A new `_ANNOTATION_FORMAT_CODEPOINTS` frozenset is unioned into
`_INVISIBLE_CODEPOINTS`, so `is_invisible_codepoint` and `sanitize_extracted_text`
strip and count them like the existing sets. No behavior change for any visible
character.

## Evidence (Required)
- **Tests:** new `tests/test_sanitize_annotation_controls.py` — all nine are
  reported invisible and stripped (with the removal count), and visible
  look-alikes (braille blank U+2800, a musical symbol) are kept.

```
$ pytest -q
490 passed, 11 skipped
$ ruff check .
All checks passed!
```

- [x] `ruff check .` is clean

Follows the file's curated-by-attack-shape design (a commented group with a
per-code-point reason). `CHANGELOG.md` untouched — git-cliff generates it from the
conventional title (#104). No open PR touches `sanitize.py`.

### DIFF
--- a/book_to_skill/sanitize.py
+++ b/book_to_skill/sanitize.py
@@ -57,10 +57,27 @@ _INVISIBLE_LETTER_CODEPOINTS = frozenset({
     0xFFA0,  # HALFWIDTH HANGUL FILLER
 })
 
+# 5. Deprecated / annotation format controls. All category Cf,
+#    Default_Ignorable, and render as nothing — the same invisible
+#    format-control shape as group 1, just blocks the original list missed.
+#    None has any use in extracted book prose.
+_ANNOTATION_FORMAT_CODEPOINTS = frozenset({
+    0x206A,  # INHIBIT SYMMETRIC SWAPPING
+    0x206B,  # ACTIVATE SYMMETRIC SWAPPING
+    0x206C,  # INHIBIT ARABIC FORM SHAPING
+    0x206D,  # ACTIVATE ARABIC FORM SHAPING
+    0x206E,  # NATIONAL DIGIT SHAPES
+    0x206F,  # NOMINAL DIGIT SHAPES
+    0xFFF9,  # INTERLINEAR ANNOTATION ANCHOR
+    0xFFFA,  # INTERLINEAR ANNOTATION SEPARATOR
+    0xFFFB,  # INTERLINEAR ANNOTATION TERMINATOR
+})
+
 _INVISIBLE_CODEPOINTS = (
     _ZERO_WIDTH_CODEPOINTS
     | _BIDI_CONTROL_CODEPOINTS
     | _INVISIBLE_LETTER_CODEPOINTS
+    | _ANNOTATION_FORMAT_CODEPOINTS
 )
 
 # 4. The Unicode tag block. Originally language tags, now used to smuggle an
--- a/tests/test_sanitize_annotation_controls.py
+++ b/tests/test_sanitize_annotation_controls.py
@@ -1,0 +1,40 @@
+"""Deprecated / interlinear-annotation format controls are stripped too.
+
+These are category Cf, Default_Ignorable code points that render as nothing —
+the same invisible format-control shape as the zero-width set — but the original
+blocklist missed them, so they passed straight through the extraction scrub (and,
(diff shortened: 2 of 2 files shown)


virgiliojr94:
Thanks for the focused hardening work. I re-ran both change sets against `master` and their respective heads.\n\nThese PRs overlap on `U+FFF9–U+FFFB`, but neither subsumes the other:\n- #178 adds variation-selector coverage (and `U+2800` handling) that #182 does not cover.\n- #182 adds `U+206A–U+206F` coverage that #178 does not cover.\n- The PRs also intentionally disagree on whether `U+2800` should be stripped.\n\nI am keeping both open while we settle the `U+2800` contract, then will consolidate the non-overlapping coverage with attribution to both contributors.

dex0shubham:
Thanks for the careful cross-check, @virgiliojr94 — agreed the two are complementary, and here's my reasoning on the U+2800 contract so you can settle it.

**I deliberately left U+2800 in.** Unlike the code points #182 adds (U+206A–U+206F and U+FFF9–U+FFFB — all category `Cf`, `Default_Ignorable`, no legitimate rendering), **U+2800 BRAILLE PATTERN BLANK is category `So`** and is the *space character of braille* — a legitimate, meaningful glyph in braille text, which this tool does extract. Stripping it unconditionally corrupts real braille content, which is a different failure mode from removing an invisible control.

So I'd frame it as two distinct problems:

- **Invisible `Cf` / Default_Ignorable controls** (both PRs) → safe to strip unconditionally; no legitimate rendering. That's #182's whole scope.
- **U+2800 as blank padding** → a *legitimate* character being *misused* in non-braille prose. That's better handled by a heuristic (e.g. flag/collapse a long run of U+2800) than a blanket strip that breaks braille books.

**Consolidation, whatever's easiest for you:**

- Merge order doesn't matter for the disjoint parts — #182 owns `U+206A–U+206F`, #178 owns the variation selectors.
- For the shared `U+FFF9–U+FFFB`: happy to **drop those three from #182** if #178 lands first (or rebase after), so there's no overlap.
- U+2800: I'd keep it out of the unconditional strip set for the braille reason above, but I'll defer to your call — glad to add a run-length heuristic in a follow-up if you want the padding case covered.

Ready to rebase/trim #182 however you decide.

virgiliojr94:
Decision: `U+2800 BRAILLE PATTERN BLANK` remains preserved; it is semantic braille content. #178 will be the consolidation vehicle because it already covers variation selectors. It will absorb your non-overlapping `U+206A–U+206F` coverage and retain the shared `U+FFF9–U+FFFB` controls, with attribution to this PR.\n\nPlease do not rebase this PR for now. I will close it as superseded after #178 has the consolidated update and passes validation.

PR #208, opened 2026-09-04
fix(deps): honor alternative parser availability

### LINKED ISSUE #207: [bug] Dependency preflight installs redundant fallback parsers
## What happened

The runtime dependency preflight treats alternative parser chains as though every
parser were required:

- PDF extraction offers to install `pdfminer.six` when `pdftotext` is absent but
  `pypdf` is already importable.
- HTML extraction offers to install `beautifulsoup4` when the preferred
  `trafilatura` parser is already importable.
- When no parser in an alternative chain is present, the preflight installs every
  alternative even though extraction only needs one.

This disagrees with `DEPENDENCY_GROUPS`, where PDF and HTML declare
`any_of_modules=True`, and can trigger unnecessary prompts and network installs.

## What you expected

For an any-of parser chain, any available parser should satisfy the preflight. If
none is available and installation is enabled, the preflight should install only
the first (preferred) parser. Formats whose dependencies are cumulative, such as
EPUB's `ebooklib` plus `beautifulsoup4`, should keep installing every missing
package.

## Source document

- Format: PDF and HTML
- Pages / size: n/a — the bug occurs before document parsing
- Language: n/a
- Does `python3 scripts/extract.py --check` show the relevant extractor installed? yes; the runtime preflight does not honor the same any-of semantics

## Repro

The failure is deterministic by stubbing dependency availability:

1. Make `python_module_available("pypdf")` return true and
   `python_module_available("pdfminer")` return false, with `pdftotext` absent.
2. Run `prepare_dependencies(".pdf", "text", "yes")`.
3. Observe an attempted `pdfminer.six` installation despite an available PDF
   parser.

The equivalent HTML case occurs with `trafilatura` available and `bs4` absent.

## Environment

- OS: platform-independent; reproduced on Windows 11 and verified under WSL/Linux
- Python version: 3.12 reproduction; full Linux suite also passes on Python 3.14
- book-to-skill version / commit: `a6cad12`

Stamina9:
## Summary (Required)

Make runtime dependency prompts honor alternative parser chains. An available PDF
or HTML parser now satisfies preflight, and a completely missing any-of chain
installs only its first preferred parser while cumulative dependency groups keep
their existing behavior.

## Type of change (Required)

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)

- **Fixes:** Runtime preflight previously interpreted PDF and HTML fallback parsers as all-required dependencies, even though `DEPENDENCY_GROUPS` and the extractors treat them as alternatives.
- **Improves:** With one usable parser installed, redundant install attempts drop from one to zero. With none installed, an any-of group installs one preferred parser instead of every alternative.
- **Adds:** Four deterministic tests covering satisfied PDF and HTML chains, preferred any-of installation, and unchanged cumulative EPUB installation.

Closes #207

Users should not be prompted to install another fallback parser when extraction can
already proceed. The runtime prompt should agree with the dependency report's
existing `any_of_modules` semantics.

`offer_dependency_install` accepts an `any_of_modules` flag. Such a group returns
as soon as any listed module is importable; if none is present, only the first
ordered module is installed and post-install verification succeeds when any module
becomes available. PDF and HTML opt into this behavior. EPUB and other cumulative
groups retain the default all-required behavior.

The module order is reused as the preference order (`pypdf` before `pdfminer`, and
`trafilatura` before `bs4`) instead of introducing a second source of parser
priority.

## Evidence (Required)

- **Tests:** `tests/test_dependency_install_semantics.py` verifies both already-satisfied chains, preferred single-parser installation, and the EPUB regression boundary.
- **Before / after:** The new targeted suite failed in three places before the implementation and passes all four tests after it. The full Linux suite and repository gates pass.

```text
Before:
3 failed, 1 passed
- attempted ['pdfminer.six'] while pypdf was available
- attempted ['beautifulsoup4'] while trafilatura was available
- offer_dependency_install had no any_of_modules semantics

After:
4 passed in 0.03s

Full Linux/WSL gate:
624 passed, 7 skipped in 2.84s
SKILL.md: no Claude Code-breaking issues (1 pre-existing soft length warning)
```

## Screenshots / output

The behavior is terminal-only; the before/after output is included in the evidence
block above.

- [x] `ruff check .` is clean

## Notes for reviewers

`SKILL.md` is unchanged. On native Windows, the current upstream Hermes Bash-path
tests have 12 pre-existing failures; the same complete suite passes under Linux/WSL
as shown above. This PR does not touch Hermes support.

### DIFF
--- a/book_to_skill/dependencies.py
+++ b/book_to_skill/dependencies.py
@@ -157,12 +157,21 @@ def offer_dependency_install(
     module_names: list[str],
     fallback: str | None,
     install_mode: str,
+    any_of_modules: bool = False,
 ) -> None:
-    packages = missing_python_packages(module_names)
-    if not packages:
+    missing_packages = missing_python_packages(module_names)
+    if not missing_packages or (
+        any_of_modules and len(missing_packages) < len(module_names)
+    ):
         return
 
-    message = f"{feature} uses {', '.join(packages)} if installed"
+    package_choices = [PYTHON_DEPENDENCIES[name] for name in module_names]
+    if any_of_modules:
+        message = f"{feature} uses one of {', '.join(package_choices)} if installed"
+        packages = missing_packages[:1]
+    else:
+        message = f"{feature} uses {', '.join(missing_packages)} if installed"
+        packages = missing_packages
     if fallback:
         message += f", otherwise {fallback}"
     message += "."
@@ -187,7 +196,12 @@ def offer_dependency_install(
 
     if install_python_packages(packages):
         still_missing = missing_python_packages(module_names)
-        if not still_missing:
+        dependencies_satisfied = (
+            len(still_missing) < len(module_names)
+            if any_of_modules
+            else not still_missing
+        )
+        if dependencies_satisfied:
             print("Package installation complete.")
             return
         print(f"Package installation incomplete; still missing: {', '.join(still_missing)}", file=sys.stderr)
@@ -213,6 +227,7 @@ def prepare_dependencies(ext: str, extraction_mode: str, install_mode: str) -> N
             module_names=["pypdf", "pdfminer"],
             fallback="any installed Python PDF parser; extraction fails if none are available",
             install_mode=install_mode,
+            any_of_modules=True,
         )
 
     if ext == ".epub":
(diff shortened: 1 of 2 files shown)


virgiliojr94 (approved):
The new tests reproduce three failures on master and pass on this head, including the cumulative-dependency boundary. Full local suite: 625 passed, 6 skipped; lint passes. No blocking finding in this diff.

PR #210, opened 2026-09-05
feat(skill): allow choosing project-local skill path

HoosseinRahimi:
## Summary

Preserve the established personal-skill default while allowing an explicit project-local destination for generated skills.

## Type of change (Required)

- [x] ✨ Feature (non-breaking change that adds capability)

## What it does (Required)

- Keeps `~/.agents/skills` as the default personal destination for non-Hermes hosts.
- Treats project-local output as an explicit choice, using a clear project request or `BOOK_TO_SKILL_SCOPE=project`.
- Supports `BOOK_TO_SKILL_SCOPE=personal` for automation without introducing a mandatory scope prompt.
- Documents that the selected destination may require host approval before writing.
- Adds a concrete scope-selection check to the installation documentation.

## Motivation & context (Required)

The original change asked every user to choose between personal and project-local output, which conflicts with the accepted default-root decision in #125. Project-local output is still useful for repository-specific, shareable skills, but it should remain opt-in. The documentation also must not claim that personal writes need no approval when the host gates writes outside the working directory.

## How it works (Required for anything non-trivial)

Step 5 resolves scope from an explicit user request or `BOOK_TO_SKILL_SCOPE`. If neither is present, it preserves the existing personal default and then applies the host-specific root rules. An explicit project-local request selects the project-local row. The documentation records the before/after behavior:

```text
Before: no scope request or BOOK_TO_SKILL_SCOPE → personal default (~/.agents/skills)
After:  BOOK_TO_SKILL_SCOPE=project or an explicit project-local request → project-local host root
```

## Evidence (Required)

- `python tools/validate_skill.py SKILL.md` — passes with one existing soft line-count warning.
- `python -m pytest tests/test_validate_skill.py tests/test_scope_selection_contract.py -q` — 12 passed.
- `ruff check --select E9,F --target-version py310 book_to_skill/ scripts/ tests/ tools/` — passed.
- `git diff --check` — passed.

## Screenshots / output (Required if behavior or output changes — Optional for pure internal refactor)

No generated output or UI changes.

## Checklist (Required — all must be checked)

- [x] One focused change (one feature/fix per PR; not a stack of unrelated work)
- [x] Branch is rebased on the latest `master` and has no merge conflicts
- [x] Tests added/updated for behavior changes
- [x] `pytest -q` is green on a clean checkout
- [x] `ruff check .` is clean
- [x] `python3 tools/validate_skill.py SKILL.md` passes (if `SKILL.md` changed)
[...]

PR #209, opened 2026-09-05
feat(hosts): add OpenClaw skill discovery and validation

HoosseinRahimi:
## Summary

Add OpenClaw support to book-to-skill, including extractor discovery for its documented skill layouts and an OpenClaw-specific validator lens.

## Type of change (Required)

- [x] ✨ Feature (non-breaking change that adds capability)

## What it does (Required)

- Adds OpenClaw to the supported-host documentation and reload guidance.
- Probes the active OpenClaw state root (flat or grouped through six grouping levels), plus flat, grouped, and deep workspace roots.
- Adds the `openclaw` validator lens with OpenClaw frontmatter keys and identifier rules.
- Adds regression coverage for flat, two-level, six-level, and custom-state extractor layouts, plus the OpenClaw lens.

## Motivation & context (Required)

OpenClaw follows the Agent Skills format and discovers skills in the active state directory's `skills/` root and workspace `skills/` locations. The shared `~/.agents/skills` compatibility root is discoverable only with the default `OPENCLAW_STATE_DIR`; a non-default state must use its active state root, a workspace root, or an explicitly configured extra directory. The previous probe only covered a flat `~/.openclaw/skills/book-to-skill` path, so documented grouped and workspace layouts could not find `scripts/extract.py`. The validator also had no OpenClaw lens.

## How it works (Required for anything non-trivial)

The probe now checks:

- `${OPENCLAW_STATE_DIR:-$HOME/.openclaw}/skills/book-to-skill/scripts/extract.py`
- `${OPENCLAW_STATE_DIR:-$HOME/.openclaw}/skills/*/.../*/book-to-skill/scripts/extract.py` (zero through six grouping directories)
- `skills/book-to-skill/scripts/extract.py`
- `skills/*/.../*/book-to-skill/scripts/extract.py` (zero through six grouping directories)
- The same workspace layouts under the enclosing Git root when invoked from a nested directory.

It does not infer arbitrary `skills.load.extraDirs`; those paths are explicitly verified with `openclaw skills list`.

The OpenClaw lens accepts its documented metadata and command-dispatch keys, does not enforce foreign `allowed-tools` restrictions, and requires lowercase hyphenated skill names.

## Evidence (Required)

- `python -m pytest tests/test_validate_skill.py tests/test_openclaw_host_support.py -q` — 12 passed, 7 skipped locally on Windows; the Bash probe cases run in the Ubuntu CI matrix.
[...]

PR #216, opened 2026-09-10
feat(hosts): add Opencode skill discovery and validation

### LINKED ISSUE #137: [feat] Allow choosing project-local skill path by default
At present, book to skill defaults to saving the generated skills to the global directory~/. claude/skills
But users usually need a lot of confirmed authorization to enable claude code writing in the project directory, which is very frustrating.
Suggest adding a generation location option:
1. Project level:
Project directory/. claude/kills
Suitable for project specific skills.

wgm66:
## Summary (Required)

Adds **OpenCode** as a first-class host for generated book skills, and fixes the converter's project-local discovery so it resolves from the **git worktree** rather than the current directory.

## Type of change (Required)

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Feature (non-breaking change that adds capability)
- [ ] Performance (faster / cheaper, no behavior change)
- [ ] Refactor (no behavior change)
- [ ] Docs only
- [ ] Security
- [ ] Chore / CI / tooling
- [ ] Breaking change (changes existing behavior or a public interface)

## What it does (Required)

- **Adds:** an OpenCode lens to `tools/validate_skill.py` (`--lens opencode`) enforcing OpenCode's real frontmatter contract — only `name`/`description`/`license`/`compatibility`/`metadata` are recognized, `allowed-tools` is not enforced (OpenCode gates tools through `permission.skill` in `opencode.json`), and the name pattern is the stricter `[a-z0-9]+(-[a-z0-9]+)*`.
- **Adds:** OpenCode to the spec's host table, personal/project skill roots, the converter discovery list, the unknown-host prompt, and the reload guidance; plus README (en/zh-CN/ru), `docs/install.md`, `docs/index.md`, `docs/architecture.md` and `docs/how-it-works.md`.
- **Fixes:** project-local discovery. The probe listed project roots as bare CWD-relative paths (`.opencode/skills/...`), so it only resolved when the agent sat at the project root. It now resolves them against the already-computed `$PROJECT_ROOT`, keeping the CWD-relative forms so non-git directories behave exactly as before.

OpenCode reads the same `SKILL.md` format but was not a named host, so the converter could place a generated skill where OpenCode never looks, and nothing validated a skill against OpenCode's stricter rules (unknown frontmatter keys are silently ignored; the name pattern is tighter than the general Agent Skills default).

Closes #137-adjacent host coverage. Verified against OpenCode's own source and docs: `~/.agents/skills` is an unconditional global root (`packages/opencode/src/skill/index.ts`), `~/.claude/skills` is a compatibility global root, `~/.config/opencode/skills` is OpenCode's own managed root, and project discovery walks up from the CWD to the git worktree. **`~/.cache/opencode/skills` is only a remote-pull cache populated from `skills.urls` — it is deliberately not documented as an authoring root.**

Instruction-level changes to `SKILL.md` plus one tooling lens — no extractor code paths. Split from the bytecode-policy work per review (that now lives in its own PR).

## Evidence (Required)

- **Tests:** `pytest tests/ -q` on Linux (Python 3.9.5) -> **698 passed, 5 skipped, 0 failed**, the 5 being pre-existing `importorskip` guards for the optional deps CI does not install (`703 passed` with them installed). The Windows dev host additionally shows 17 `test_hermes_host_support.py` failures caused solely by a bare `bash` resolving to the WSL stub - that file passes 20/20 on Linux. `ruff check .` clean; `validate_skill.py SKILL.md` exit 0.
- **Nested-directory regression locked in:** `tests/test_opencode_host_support.py` now runs **every** layout from `project/src/nested` (and a new case from `project/src/nested/deeper` over `.opencode`/`.agents`/`.claude`). Negative control: reverting the `$PROJECT_ROOT` block fails exactly the 4 nested/project cases (`4 failed, 12 passed`); restoring → green.
- **Policy contract:** `tests/test_scope_selection_contract.py` gains checks that the merged selection policy is intact — no merge residue, exactly one `Selection rules:` list preserving #125's `~/.agents/skills` personal default, no private-root preference for OpenCode, and the OpenCode download cache is never presented as a destination.

### Linux verification (independent of the Windows dev host)

Run on a clean **Ubuntu 20.04 / Python 3.9.5** guest - the lowest bound of this
repo's CI matrix (`python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]`,
`runs-on: ubuntu-latest`) - on a fresh checkout of this head.

**CI parity** - a venv with only `pytest`, exactly as `ci.yml` installs it:

```
pytest tests/ -q                 -> 698 passed, 5 skipped, 0 failed
ruff check --select E9,F --target-version py310 book_to_skill/ scripts/ tests/ tools/
                                 -> All checks passed!
tools/validate_skill.py SKILL.md -> exit 0
smoke (dependency-free extract)  -> scripts/extract.py exit 0, full_text.txt +
                                    metadata.json produced, "Backpressure" present
```

The 5 skips are pre-existing `importorskip` guards for optional deps that
`ci.yml` deliberately does not install (`bs4` in `test_html_block_boundaries.py`,
`trafilatura` in `test_html_boilerplate_extraction.py`) - the same 5 that skip on
Windows. With those two deps installed the same tree gives
**703 passed, 0 skipped**, so nothing here is hidden by a skip.

This matters because the Windows dev host reports 17 failures in
`tests/test_hermes_host_support.py`, and those are an environment artifact rather
than a defect: a bare `bash` there is resolved by `CreateProcess` against the
system directory before `PATH`, so it picks the WSL stub. On Linux
`shutil.which("bash")` is `/usr/bin/bash`, and **that same file passes 20/20**.

- [x] `ruff check .` is clean

**Rebased onto current `master` (`f08d343`)** and the Step 5 conflict from #210 resolved so the scope paragraph and #125's six selection rules both survive.

**Merge residue removed.** Step 5 previously carried two literal `HEAD` lines and a duplicated pre-#125 "ask on a fresh machine" rule list, left over from an earlier conflict resolution. Both are gone in every commit of the series — verified per-commit — and the new contract test fails if either returns.

**Systematic root cause, not just a one-off.** The residue existed because my conflict resolution had kept both sides of a conflict. The appended list is the specific defect; the guard is the new contract test.

**Test-infrastructure fix included.** These tests had been silently skipping on Windows: a bare `bash` is resolved by `CreateProcess` against the system directory before `PATH`, so it always picked the WSL stub (`C:\Windows\System32\bash.exe`, exit 255) even with a working Git bash installed. Detection now validates absolute candidates with `bash -c true`, and MSYS paths are normalized through `cygpath -w` — so the suite actually runs instead of skipping.

Scope is OpenCode host support + the nested-discovery fix only; the bytecode/cache-cleanup policy is a separate PR.


virgiliojr94 (changes requested):
From `project/src/nested`, the probe cannot find `project/.opencode/skills/book-to-skill/scripts/extract.py`. Moving the test to the project root hides that supported discovery case; please resolve the project root and keep the nested-directory regression. Also split the global bytecode policy/cache-cleanup changes into their own PR and update the description to match the final scope.

virgiliojr94 (changes requested):
Nested-directory discovery is fixed and the bytecode work is split out; 662 tests pass, with 6 skips. Step 5 still contains two literal HEAD lines and both the new personal-default rules and the old ask-on-fresh-install rules. Remove the merge residue and keep one consistent selection policy preserving #125; the new instruction to prefer OpenCode's private root also contradicts that default. Add a contract check for the resolved policy, since the current green tests miss this regression.

wgm66:
Addressed in `7573dcb` (rebased onto `f08d343`). Point by point:

**Two literal `HEAD` lines + the duplicated rule list — removed.** You were right that I'd left both the new and the old policy in place: `SKILL.md` had a literal `HEAD` inside the host table, a second one above the rule list, and a second numbered list (`1.`–`4.`, the pre-#125 "ask on a fresh machine" wording) sitting directly under #125's six rules. All three are gone, and I verified it **per commit**, not just at the tip, since the series is reviewed commit-by-commit: every commit now reports 0 `HEAD`, 0 `fresh machine`, exactly 1 `Selection rules:` block.

**One consistent policy preserving #125.** Step 5 now carries a single 6-rule list; rule 1 still sets `SKILLS_HOME` to `~/.agents/skills`, and the #210 scope paragraph (`First resolve **scope** … preserve the established personal default`) is intact — the rebase conflict was resolved to keep both #210 and #125.

**The OpenCode private-root preference is gone — and you were right about the facts.** I checked OpenCode's source rather than my own notes: `~/.agents/skills` is an unconditional global root (`packages/opencode/src/skill/index.ts`, alongside the `~/.claude/skills` compatibility root), and `~/.config/opencode/skills` is its own managed root. So #125's default already serves OpenCode and there was no reason to prefer the private root. The table row and prose now say the cross-agent root is the default and the OpenCode-managed root is opt-in. Related: `~/.cache/opencode/skills` is only a remote-pull cache populated from `skills.urls`, so it is deliberately **not** listed as an authoring root.

**Contract check added.** `tests/test_scope_selection_contract.py` now locks the resolved policy: no residue/markers, exactly one contiguous 1..6 rule list, the pre-#125 wording absent, `~/.agents/skills` still the personal default, the OpenCode row ordering the cross-agent root first, no private-root preference, and the cache root never a destination. I proved each check actually fires by re-introducing the defect: literal `HEAD` → `AssertionError: literal 'HEAD' merge residue on line(s) [376]`; re-adding the old list → `Selection rules: must be one contiguous 1..6 list, got []`; private-root-first row → `OpenCode row must present ~/.agents/skills before the private root`.

**Also fixed the same inconsistency in the docs I added** — `docs/install.md`, `docs/how-it-works.md` and the three READMEs presented the OpenCode private root as the primary destination; they now match the policy.

`pytest -q` → 677 passed, 9 skipped (same 17 pre-existing WSL-only Windows failures); `ruff check .` clean; `validate_skill.py SKILL.md` exit 0.

wgm66:
Follow-up: I re-ran everything on **real Linux** instead of the Windows dev host, because the Windows numbers needed a caveat and I did not want to leave that ambiguity standing.

**Environment:** clean Ubuntu 20.04 guest (VirtualBox), **Python 3.9.5** — the lowest bound of this repo's CI matrix — pytest 8.4.2, ruff 0.16.8, on a fresh checkout of this head with the optional deps installed so nothing skips.

```
pytest -q                        -> 703 passed, 0 failed, 0 skipped
ruff check .                     -> All checks passed!
tools/validate_skill.py SKILL.md -> exit 0
```

**The caveat is now resolved rather than excused.** The 17 failures I previously reported as "pre-existing on Windows" are confirmed to be purely an environment artifact: on Linux `shutil.which("bash")` resolves to `/usr/bin/bash`, and `tests/test_hermes_host_support.py` passes **20/20**. On Windows that same file picks up the WSL stub via `CreateProcess`'s system-directory-first lookup, which is the entire cause. Nothing was being masked.

**Behaviour this PR changes, reproduced on Linux directly (not only via pytest):**

```
nested-CWD probe, run from project/src/nested with real /usr/bin/bash:
  .opencode -> project/.opencode/skills/book-to-skill/scripts/extract.py  OK
  .agents   -> project/.agents/skills/book-to-skill/scripts/extract.py    OK
  .claude   -> project/.claude/skills/book-to-skill/scripts/extract.py    OK
```

**Policy contract, on Linux:** `tests/test_opencode_host_support.py` +
`tests/test_scope_selection_contract.py` + `tests/test_hermes_host_support.py`
-> **45 passed**.

Full command output is reproducible from the guest at `~/bts/pr216`.

virgiliojr94 (changes requested):
The previous policy/residue issues are fixed; this head passes 697 tests (6 skipped), Ruff and skill validation. Integration is now blocked by conflicts with merged #209 in eight files. Rebase while preserving both hosts and OpenClaw custom-state rules. Update test_skill_has_exactly_one_selection_rule_list: it hard-codes exactly six rules and the old host prompt, but master legitimately has eight after #209. Verify a single contiguous list and the required policy instead of freezing its length; rerun both host probe suites on the resolved tree.

PR #218, opened 2026-09-12
fix(pdf): fall back on invalid inspector metadata

### LINKED ISSUE #217: [bug] Invalid pdf-inspector metadata can abort batch extraction
## What happened

`inspect_pdf()` catches exceptions raised by `pdf_inspector.process_pdf()`, but
normalises the returned object outside that protected block. If an optional
pdf-inspector result contains an unexpected value or shape, conversions such as
`float(result.confidence)`, `int(result.page_count)`, or `list(...)` raise
`ValueError`/`TypeError`.

The exception then escapes the pdf-inspector hook instead of allowing the
existing PDF extraction chain to run. In a multi-source invocation it can also
bypass the batch loop, which catches `ExtractionError`, and stop processing the
remaining sources.

This is a defensive compatibility issue at the optional dependency boundary;
the supported pdf-inspector API currently documents the expected field types.

## What you expected

Any failure while consuming the optional preflight result should produce the
existing warning and return `(None, None)`. The hook should then delegate to the
existing Docling/pdftotext/pypdf/pdfminer extraction chain, preserving graceful
degradation.

## Source document
- Format: PDF (a synthetic `%PDF-1.7` fixture is sufficient)
- Pages / size: one minimal fixture; content is not relevant
- Language: N/A
- Does `python3 scripts/extract.py --check` show the relevant extractor installed? yes; the deterministic regression uses a stubbed module result

## Repro

On current `master` (`0ad050e`), make `pdf_inspector.process_pdf()` return a
result whose `confidence` is `"not-a-number"`, then call
`inspect_pdf("book.pdf")`.

Observed:

```text
ValueError: could not convert string to float: 'not-a-number'
```

Equivalent uncaught failures occur for a non-integer `page_count` and a
non-iterable list field such as `pages_needing_ocr`.

## Environment
- OS: Windows 11 for reproduction; behavior is platform-independent
- Python version: 3.12.10
- book-to-skill version / commit: `0ad050e`

Stamina9:
## Summary (Required)

Make the optional pdf-inspector preflight fail closed when its returned metadata
cannot be normalised. Invalid result fields now trigger the existing warning and
delegate to Book-to-Skill's established PDF extraction chain instead of escaping
as `ValueError` or `TypeError`.

## Type of change (Required)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Fixes:** Metadata conversions performed after `process_pdf()` were outside the adapter's exception boundary, so malformed or incompatible optional-dependency results could abort extraction rather than fall back.
- **Adds:** Regression coverage for invalid confidence, page count, and iterable fields, plus an end-to-end hook assertion that the original extraction function is called.

pdf-inspector is an optional preflight. Its failure should never take authority
away from the existing extractor, consistent with the project's graceful
degradation design. Protecting both the library call and result normalisation
keeps that boundary intact without changing successful pdf-inspector behavior.

Closes #217

The existing `try` block now covers both `process_pdf()` and all reads and
normalisation of its returned object. Any ordinary exception follows the already
established warning path and returns `(None, None)`; the hook therefore invokes
the original extractor. Valid metadata and trusted native Markdown follow the
same path as before.

The change deliberately does not alter confidence thresholds, PDF routing,
metadata shape, or the legacy extraction implementation.

## Evidence (Required)
- **Tests:** Added three parameterised adapter cases (`confidence`, `page_count`, and `pages_needing_ocr`) and one hook-level fallback case.
- **Before / after:** The new tests fail on current master with uncaught `ValueError`/`TypeError`; all pass after the exception boundary is extended.

```text
# Before
4 failed, 5 passed in 0.11s

# After
9 passed in 0.05s

# Full Linux/WSL gate from a clean `git archive` of the commit
633 passed, 8 skipped in 1.18s
```

- [x] `ruff check .` is clean

None. Valid pdf-inspector results retain their existing behavior; only failures
while consuming an optional result now use the existing fallback.

The test uses deliberately malformed stub results to exercise the adapter
boundary. The supported pdf-inspector API documents typed fields; this PR does
not claim that a specific released version currently returns malformed values.

### DIFF
--- a/book_to_skill/pdf_inspector_integration.py
+++ b/book_to_skill/pdf_inspector_integration.py
@@ -58,6 +58,41 @@ def inspect_pdf(path: str | Path) -> tuple[str | None, dict[str, Any] | None]:
 
     try:
         result = pdf_inspector.process_pdf(str(path))
+        pdf_type = _normalise_pdf_type(getattr(result, "pdf_type", None))
+        confidence = float(getattr(result, "confidence", 0.0) or 0.0)
+        pages_needing_ocr = list(getattr(result, "pages_needing_ocr", None) or [])
+        has_encoding_issues = bool(getattr(result, "has_encoding_issues", False))
+        markdown = getattr(result, "markdown", None)
+
+        native_markdown_trusted = bool(
+            isinstance(markdown, str)
+            and markdown.strip()
+            and pdf_type == "text_based"
+            and confidence >= _MIN_NATIVE_CONFIDENCE
+            and not pages_needing_ocr
+            and not has_encoding_issues
+        )
+
+        metadata: dict[str, Any] = {
+            "engine": "pdf-inspector",
+            "version": _package_version(),
+            "pdf_type": pdf_type,
+            "confidence": round(confidence, 4),
+            "native_markdown_trusted": native_markdown_trusted,
+            "pages_needing_ocr": pages_needing_ocr,
+            "ocr_reasons_by_page": _ocr_reasons(
+                getattr(result, "ocr_reasons_by_page", None)
+            ),
+            "pages_with_tables": list(
+                getattr(result, "pages_with_tables", None) or []
+            ),
+            "pages_with_columns": list(
+                getattr(result, "pages_with_columns", None) or []
+            ),
+            "has_encoding_issues": has_encoding_issues,
+            "is_complex_layout": bool(getattr(result, "is_complex_layout", False)),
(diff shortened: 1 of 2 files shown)


virgiliojr94 (approved):
The four new regression cases fail on master and all nine adapter tests pass on this head. Extending the existing exception boundary preserves the fallback chain without changing successful results. No blocking finding in this diff.

PR #220, opened 2026-09-12
fix(pdf): discard stale inspection metadata

### LINKED ISSUE #219: [bug] Failed pdf-inspector reinspection can reuse stale metadata
## What happened

The pdf-inspector integration keeps successful inspection results in the
process-level `_INSPECTIONS` mapping so they can later be added to
`metadata.json`. An entry is overwritten when a later inspection succeeds, but
it is not invalidated before a new attempt.

When the same input path is processed again in the same Python process and the
new pdf-inspector preflight returns no metadata (for example, it is unavailable
or fails), the previous result remains cached. The next metadata enrichment can
therefore report stale pdf-inspector provenance for a run in which that
inspection did not succeed.

This affects repeated programmatic/embedded CLI use; separate command-line
processes do not share the mapping.

## What you expected

Each extraction attempt should replace the cached state for its own input path.
If the current preflight produces no inspection metadata, enrichment for that
path should contain no `pdf_inspector` entry. Results for other files in the
same batch should remain intact.

## Source document
- Format: PDF (synthetic `%PDF-1.7` fixture)
- Pages / size: minimal fixture; document content is not relevant
- Language: N/A
- Does `python3 scripts/extract.py --check` show the relevant extractor installed? the deterministic reproduction stubs the optional preflight result

## Repro

1. Install the pdf-inspector hook once in a Python process.
2. Process a PDF path with an inspection result containing `confidence: 0.5`.
3. Process the same path again, with `inspect_pdf()` returning `(None, None)`.
4. Enrich a fresh `metadata.json` for the second run.

On `master` (`01f8a74`), the second metadata file incorrectly contains the
first run's pdf-inspector result:

```json
{"pdf_inspector": {"engine": "pdf-inspector", "confidence": 0.5}}
```

## Environment
- OS: Windows 11 for reproduction; state behavior is platform-independent
- Python version: 3.12.10
- book-to-skill version / commit: `01f8a74`

Stamina9:
## Summary (Required)

Invalidate a path's cached pdf-inspector result before every new extraction
attempt. A failed or unavailable reinspection can no longer leak an earlier
run's provenance into a fresh `metadata.json`.

## Type of change (Required)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Fixes:** The process-level inspection cache retained metadata for a path when a later preflight for that same path returned no result.
- **Adds:** A repeated-run regression test proving that fresh output is not enriched with stale pdf-inspector provenance.

`_INSPECTIONS` bridges extraction and the later metadata-enrichment step, but it
persists for the lifetime of the Python process. Embedded callers and tests can
invoke the pipeline more than once. Metadata should describe the current run,
not the last successful inspection of the same path.

Closes #219

The wrapper computes the resolved input key once and removes only that key
before deciding whether or how to inspect the file. It writes the key back only
when the current `inspect_pdf()` call returns metadata. Per-file invalidation
preserves inspection results already collected for other sources in the same
batch.

No inspection schema, routing threshold, extractor behavior, or public API is
changed.

## Evidence (Required)
- **Tests:** Added `test_failed_reinspection_does_not_reuse_stale_metadata`, which runs the same path twice, makes the second preflight return no result, enriches fresh metadata, and asserts that neither metadata level receives a stale `pdf_inspector` entry.
- **Before / after:** The new regression fails on merged `master` because the first result is reused; it passes after per-path invalidation.

```text
# Before
1 failed, 9 passed in 0.08s

# After (targeted)
10 passed in 0.04s

# Full clean Linux gate
634 passed, 8 skipped in 1.21s
```

- [x] `ruff check .` is clean

None. Current-run inspection results are still recorded and enriched exactly as
before; only stale results from an earlier attempt are discarded.

The cache is invalidated per resolved path rather than globally so processing a
later source does not remove provenance already collected for other PDFs in the
same batch.

### DIFF
--- a/book_to_skill/pdf_inspector_integration.py
+++ b/book_to_skill/pdf_inspector_integration.py
@@ -191,12 +191,15 @@ def install_pdf_inspector_hook(utils_module: Any | None = None) -> None:
         return
 
     def wrapped(input_path: Path, extraction_mode: str, install_mode: str) -> dict[str, Any]:
+        inspection_key = str(input_path.resolve())
+        _INSPECTIONS.pop(inspection_key, None)
+
         if not _looks_like_pdf(input_path):
             return original(input_path, extraction_mode, install_mode)
 
         markdown, inspection = inspect_pdf(input_path)
         if inspection is not None:
-            _INSPECTIONS[str(input_path.resolve())] = inspection
+            _INSPECTIONS[inspection_key] = inspection
 
         if extraction_mode == "text" and markdown and inspection:
             result = _result_from_inspector(utils_module, input_path, markdown, inspection)
--- a/tests/test_pdf_inspector_integration.py
+++ b/tests/test_pdf_inspector_integration.py
@@ -102,6 +102,43 @@ def test_hook_uses_existing_pipeline_when_inspector_metadata_is_invalid(
     assert integration._INSPECTIONS == {}
 
 
+def test_failed_reinspection_does_not_reuse_stale_metadata(tmp_path, monkeypatch):
+    integration._reset_state_for_tests()
+    pdf = tmp_path / "book.pdf"
+    pdf.write_bytes(b"%PDF-1.7\nfixture")
+    inspections = iter(
+        [
+            (None, {"engine": "pdf-inspector", "confidence": 0.5}),
+            (None, None),
+        ]
+    )
+    monkeypatch.setattr(integration, "inspect_pdf", lambda _path: next(inspections))
+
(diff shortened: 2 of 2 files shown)


virgiliojr94 (approved):
Review findings addressed; validated against current master `a45a2c9`. Full pytest, Ruff and skill validation pass on the integration tree. The new regression fails on the baseline and passes here.

PR #222, opened 2026-09-12
chore(skill): keep __pycache__ out of deployed skill directories

wgm66:
## Summary (Required)

Keeps `__pycache__` out of a deployed skill directory by setting the bytecode policy **in the executable entry points**, not in a library import.

## Type of change (Required)

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Feature (non-breaking change that adds capability)
- [ ] Performance (faster / cheaper, no behavior change)
- [x] Refactor (no behavior change)
- [ ] Docs only
- [ ] Security
- [x] Chore / CI / tooling
- [ ] Breaking change (changes existing behavior or a public interface)

## What it does (Required)

- **Adds:** `if __name__ == "__main__": sys.dont_write_bytecode = True` before the package import in each module that is genuinely run as a script — `scripts/extract.py`, `tools/scan_generated_skill.py`, `tools/discovery_tax.py`, `book_to_skill/cli.py`. `__name__` is bound before the module body runs, so the policy applies to a direct invocation and never mutates an embedding process that merely imports the module.
- **Removes:** the import-time `sys.dont_write_bytecode = True` from `book_to_skill/utils.py` (a pure library module with no `__main__` guard) and the unconditional import-time assignment in `book_to_skill/cli.py`.
- **Rewrites:** `tests/test_no_bytecode_pollution.py` so it proves the invocation worked and measures artifacts in an **isolated deployment copy** instead of cleaning the developer's source tree.

When book-to-skill is installed as an agent skill, the skill *is* the source tree — it is cloned or unpacked into `~/.config/opencode/skills/<name>/` (or another host root) and run from there. Any `__pycache__` written on first run ships inside the skill: extra files to scan, noise in diffs, and `__pycache__/*.pyc` wherever the skill is published or shared. The agent invoking the skill does not set `PYTHONDONTWRITEBYTECODE`, so the guard has to live in the entry points themselves.

## Evidence (Required)

- **The regression test can now fail.** With a deliberately broken entry point (`scripts/extract.py` → `raise SystemExit(1)`), the suite reports **3 failed** (`test_entry_point_writes_no_bytecode[scripts/extract.py]`, `test_entry_points_set_the_policy_before_importing_the_package`, `test_entry_point_list_still_matches_the_source_tree`); restoring the file → **8 passed**. A dedicated negative-control test asserts the helper raises rather than passing on a discarded return code.
- **Artifacts measured in isolation.** The old `_clean()` deleted `__pycache__` from the developer's checkout; the new `deployed` fixture copies the payload into `tmp_path` and measures only there.
- **Measured on upstream before/after:** running each entry point with `PYTHONDONTWRITEBYTECODE` stripped wrote 14–16 `.pyc` files; after the guard, all four run `--help` with exit 0 and **0 `.pyc` / 0 `__pycache__`** in a real isolated deployment.
- **Library import is inert:** `sys.dont_write_bytecode=False; import book_to_skill, book_to_skill.utils, book_to_skill.cli` → still `False`.
- **Tests:** `pytest tests/ -q` on Linux (Python 3.9.5) -> **673 passed, 5 skipped, 0 failed**, the 5 being pre-existing `importorskip` guards for the optional deps CI does not install (`678 passed` with them installed). The Windows dev host additionally shows 17 `test_hermes_host_support.py` failures caused solely by a bare `bash` resolving to the WSL stub - that file passes 20/20 on Linux. `ruff check .` clean; `validate_skill.py SKILL.md` exit 0.

### Linux verification (independent of the Windows dev host)

Run on a clean **Ubuntu 20.04 / Python 3.9.5** guest - the lowest bound of this
repo's CI matrix (`python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]`,
`runs-on: ubuntu-latest`) - on a fresh checkout of this head.

**CI parity** - a venv with only `pytest`, exactly as `ci.yml` installs it:

```
pytest tests/ -q                 -> 673 passed, 5 skipped, 0 failed
ruff check --select E9,F --target-version py310 book_to_skill/ scripts/ tests/ tools/
                                 -> All checks passed!
tools/validate_skill.py SKILL.md -> exit 0
smoke (dependency-free extract)  -> scripts/extract.py exit 0, full_text.txt +
                                    metadata.json produced, "Backpressure" present
```

The 5 skips are pre-existing `importorskip` guards for optional deps that
`ci.yml` deliberately does not install (`bs4` in `test_html_block_boundaries.py`,
`trafilatura` in `test_html_boilerplate_extraction.py`) - the same 5 that skip on
Windows. With those two deps installed the same tree gives
**678 passed, 0 skipped**, so nothing here is hidden by a skip.

This matters because the Windows dev host reports 17 failures in
`tests/test_hermes_host_support.py`, and those are an environment artifact rather
than a defect: a bare `bash` there is resolved by `CreateProcess` against the
system directory before `PATH`, so it picks the WSL stub. On Linux
`shutil.which("bash")` is `/usr/bin/bash`, and **that same file passes 20/20**.

- [x] `ruff check .` is clean
- [x] `python3 tools/validate_skill.py SKILL.md` passes (N/A — this PR does not touch `SKILL.md`)

**Review response.** Two things changed after the first review:

1. **The policy no longer leaks out of a library import.** It was set in `book_to_skill/utils.py` at module scope, which mutates `sys.dont_write_bytecode` for the whole embedding process. It now lives inside the `__main__` guard of the entry points that are actually executed, before the package import.
2. **The test can no longer pass vacuously.** The old helper discarded the subprocess return code, so an entry point that immediately raised `SystemExit(1)` still produced a pass — the maintainer's exact reproduction. It now asserts both a zero return code and the expected output, and the negative control above proves the assertion fires.

Known limit, stated rather than hidden: `python -m book_to_skill` cannot be guarded from inside the package, because Python compiles `book_to_skill/__init__.py` before any of our code runs. `scripts/extract.py` is the documented path and is guarded.

The four guarded modules are the real script entry points only; `book_to_skill/utils.py` is a library module and is intentionally absent from `ENTRY_POINTS`. The list is kept honest by `test_entry_point_list_still_matches_the_source_tree`, which fails if a new entry point appears.


virgiliojr94 (changes requested):
Importing book_to_skill now changes sys.dont_write_bytecode for the entire embedding process via utils.py; please keep the policy at executable entry points rather than library import. The regression also passes when scripts/extract.py immediately raises SystemExit(1): its return code is discarded, and a separate import cannot prove that invocation worked. Assert successful execution/output in the tested subprocess and measure artifacts in an isolated deployment copy instead of cleaning the developer's source tree. Local reproduction: the deliberately broken entry point still gives 1 passed.

wgm66:
Addressed in `efc90bb` (rebased onto `f08d343`). Point by point:

**Policy moved off the library import — your first objection.** `book_to_skill/utils.py` is a pure library module (no `__main__` guard, imported by `__init__`, `cli` and the tests), so setting `sys.dont_write_bytecode` at module scope leaked into the whole embedding process. That assignment is deleted, as is the import-time one in `book_to_skill/cli.py`. The policy now lives inside the `__main__` guard of the modules that are actually executed — `scripts/extract.py`, `tools/scan_generated_skill.py`, `tools/discovery_tax.py`, `book_to_skill/cli.py` — placed before the package import. `__name__` is bound before the module body runs, so a direct invocation is protected and an embedder that merely imports the package is not. Verified: `sys.dont_write_bytecode=False; import book_to_skill, book_to_skill.utils, book_to_skill.cli` → still `False`.

**Return code no longer discarded — your reproduction now fails.** The helper asserted only "no artifacts", so `SystemExit(1)` passed. It now asserts a zero return code **and** the expected output. Reproducing your case exactly: replacing `scripts/extract.py` with `raise SystemExit(1)` gives **3 failed** (`test_entry_point_writes_no_bytecode[scripts/extract.py]`, `test_entry_points_set_the_policy_before_importing_the_package`, `test_entry_point_list_still_matches_the_source_tree`) instead of `1 passed`; restoring the file → 8 passed. There is also a dedicated negative-control test asserting the helper raises rather than passing.

**Artifacts measured in an isolated deployment copy.** The old `_clean()` deleted `__pycache__` out of the developer's checkout. The new `deployed` fixture copies the payload into `tmp_path` and measures only there, so the test cannot touch the source tree.

Measured before/after on upstream with `PYTHONDONTWRITEBYTECODE` stripped: each entry point wrote 14–16 `.pyc`; now all four run `--help` with exit 0 and 0 `.pyc` / 0 `__pycache__` in a real isolated deployment.

One limit I'm stating rather than hiding: `python -m book_to_skill` cannot be guarded from inside the package, because Python compiles `book_to_skill/__init__.py` before any of our code runs. `scripts/extract.py` is the documented path and is guarded.

`pytest -q` → 652 passed, 9 skipped (same 17 pre-existing WSL-only Windows failures); `ruff check .` clean.

wgm66:
Follow-up: I re-ran everything on **real Linux** instead of the Windows dev host, because the Windows numbers needed a caveat and I did not want to leave that ambiguity standing.

**Environment:** clean Ubuntu 20.04 guest (VirtualBox), **Python 3.9.5** — the lowest bound of this repo's CI matrix — pytest 8.4.2, ruff 0.16.8, on a fresh checkout of this head with the optional deps installed so nothing skips.

```
pytest -q                        -> 678 passed, 0 failed, 0 skipped
ruff check .                     -> All checks passed!
tools/validate_skill.py SKILL.md -> exit 0
```

**The caveat is now resolved rather than excused.** The 17 failures I previously reported as "pre-existing on Windows" are confirmed to be purely an environment artifact: on Linux `shutil.which("bash")` resolves to `/usr/bin/bash`, and `tests/test_hermes_host_support.py` passes **20/20**. On Windows that same file picks up the WSL stub via `CreateProcess`'s system-directory-first lookup, which is the entire cause. Nothing was being masked.

**Bytecode policy, reproduced on Linux directly (not only via pytest):**

```
isolated deployment copy, all four entry points run with PYTHONDONTWRITEBYTECODE stripped:
  scripts/extract.py             exit=0
  tools/scan_generated_skill.py  exit=0
  tools/discovery_tax.py         exit=0
  book_to_skill/cli.py           exit=0
  RESULT pyc=0 __pycache__=0 -> CLEAN
  import book_to_skill ... -> sys.dont_write_bytecode == False

negative control (scripts/extract.py -> raise SystemExit(1)):
  3 failed  (test_entry_point_writes_no_bytecode[scripts/extract.py],
             test_entry_points_set_the_policy_before_importing_the_package,
             test_entry_point_list_still_matches_the_source_tree)
restored:
  8 passed
```

Full command output is reproducible from the guest at `~/bts/pr222`.

virgiliojr94 (approved):
Validated against current master `526f362`: full pytest, Ruff, skill validation and synthetic CLI smoke pass on the integration tree. The new regressions fail on the baseline and pass here.

PR #224, opened 2026-09-12
fix(skill): stop and ask before re-converting an existing skill (re-run guard)

Scan-law4:
## Summary (Required)

Adds a re-run guard to Step 0 of the converter spec: before a Full Conversion starts extraction, it derives the prospective skill slug and, if that skill already exists, stops and asks the user (fold in / verify / force) instead of silently re-extracting. Interrupted-run recoveries resume the existing work directory only when the workdir, its manifest, and the extraction mode still match the current inputs; otherwise a fresh extraction is started.

## Type of change (Required)

## What it does (Required)

- **Fixes:** a Full Conversion of a source whose target skill already exists re-runs the entire extraction (Steps 2–3) before Step 5 derives the skill slug and notices the collision. Mode 4 (Update/Fold-in) cannot catch a naive re-run: its trigger requires the user to explicitly request an update, and the existing-skill check in Step 0 only fires when the *input path itself* is a skill directory or the user supplied the slug as an argument — neither holds when the user re-issues the original conversion command. Most visible on session recovery, where an interrupted run gets re-driven.
- **Improves:** re-invoking an already-converted source now costs one slug check instead of a full re-extraction. In our use case a ~145-page PDF was fully extracted three times in one working session ([measured account](https://gist.github.com/Scan-law4/c5f3e578d6289c10ef45778d3d38dbc0)).
- **Adds:** a user-facing stop-and-choose prompt on slug collision (fold in via Mode 4, verify-and-stop, or force regeneration) and a conditional-resume rule for recovery after interrupted runs (fresh extraction when workdir, manifest, or mode mismatch).

Closes #n/a — no existing issue tracks this; the re-run/recovery gap surfaced during a long working session with network interruptions (account linked above, with measurements).

One bullet added to **Step 0** (an existing step, per the lean-SKILL.md rule — no new steps, no code changes, no change to generated-skill behavior):

> **Re-run guard.** Before starting extraction for a Full Conversion, derive the prospective skill slug under Step 5's naming options — the author-concept form, the by-title form, and any `SKILL_NAME` given — and check `SKILLS_HOME` for an existing match. On a match, STOP and ask the user: "`{skill-name}` already exists. Choose: (1) Update/Fold-in (Mode 4), (2) verify the existing skill is complete and stop, or (3) force full regeneration." Do not re-extract until the user chooses. On context recovery after an interrupted run, reuse the existing extraction work directory **only if** the directory exists, its `metadata.json` is readable, the `sources` it lists match the files being converted now (by filename **and size**), and its recorded `extraction_mode` matches the mode this run is using; otherwise say so and start a fresh extraction instead of resuming.

**Both Step 5 naming forms are checked on purpose:** Step 5 can name a generated skill either by author-concept or by title, so a title-only lookup misses an existing author-concept skill (in our case the book *Generative AI and Agentic Operations* produced `chilukuri-agenticops` — a title-based check alone would not have found it). A test pins this.

The guard deliberately reuses Mode 4's terminology and the Step 5 naming rules rather than introducing new concepts. Alternative rejected: enforcing this in `scripts/extract.py` (e.g. a `--force` flag) — happy to move it there if maintainers prefer code over spec prose, but the extraction step is host-orchestrated, so the decision point lives in the spec.

## Evidence (Required)

- **Tests:** new `tests/test_rerun_guard.py` (same SKILL.md-contract pattern as `test_hermes_host_support.py`): asserts the guard text exists inside Step 0, precedes the extraction step, and covers **both** Step 5 naming forms (author-concept and by-title). The test suite now includes 8 assertions: gated reuse (with the old unconditional wording asserted gone), source+config match coverage, missing/stale-workdir fresh-extraction fallback, resolution-before-lookup ordering, and a Step-0-has-no-`SKILLS_HOME` contract.
- **Before / after:** baseline on clean master: `pytest -q` → 636 passed, 5 skipped. With this PR: 644 passed, 5 skipped. `ruff check .` clean on both. `python3 tools/validate_skill.py SKILL.md` → ✓ no breaking issues (the >500-line soft warning is pre-existing: master is at 817 lines).

```
$ pytest -q
...................................................................      [100%]
644 passed, 5 skipped in 1.38s
$ ruff check .
All checks passed!
  WARN  body: 818 lines > 500 (soft guideline for optimal performance)
```

Behavior change is in the host-agent conversation flow (stop-and-ask instead of silent re-extraction); not screenshotable here. The measured before/after account of the underlying re-run problem is in the linked use-case Gist.

- [x] `ruff check .` is clean

- `SKILL.md` grows by exactly one line (817 → 818); the justification is the measured re-run cost above.
- If maintainers would rather enforce the guard in `scripts/extract.py`, I'm glad to rework it — the use-case evidence carries over either way.


virgiliojr94 (changes requested):
The recovery rule says to reuse the workdir and never restart extraction, but a deleted/cleaned workdir or changed input requires a new extraction; blindly reusing it can select stale source data. Gate reuse on an existing workdir matching the current source/config, with a clear fresh-extraction fallback. Also resolve the destination before the Step 0 lookup: SKILLS_HOME is only selected in Step 5, and a project-local request must not check the personal root by accident. The three tests only check wording; please add before/after scenarios for missing/stale workdirs and an explicit project destination.

Scan-law4:
Thanks for the review — all three points are addressed on the branch (`707e47e`).

1. **Workdir reuse is now conditional.** Reuse requires an intact, matching manifest: the
   work directory exists, its `metadata.json` is readable, the `sources` it lists are the
   files being converted now with the same filenames **and sizes**, and its recorded
   `extraction_mode` is the mode this run is using. A missing work directory (temp
   cleanup), a missing manifest, changed sources or sizes, or a different mode means:
   say so and start a fresh extraction instead of resuming — and when in doubt, ask
   before discarding or resuming. The comparison is against `metadata.json`, which
   already records `workdir`, `total_sources`, `extraction_mode` and a per-source list
   with `filename` / `source_file` / `file_size_mb`, so the check is against data the
   run actually wrote rather than an inferred state.

2. **Destination resolved before the lookup.** Step 0 now opens with the destination
   root — personal or project-local for the user's host, per the Step 5 table and
   selection rules — and both the re-run guard and the Mode 4 detection check that
   resolved root. Step 0 no longer references `SKILLS_HOME` at all, since it is only
   selected in Step 5; a project-local request can no longer consult the personal root
   by accident.

3. **Tests.** 3 → 8 in `tests/test_rerun_guard.py`: gated reuse (with the old
   unconditional wording asserted gone), source+config match coverage, the
   missing/stale-workdir fresh-extraction fallback, resolution-before-lookup ordering,
   and a Step-0-has-no-`SKILLS_HOME` contract. To be precise about what they are: these
   assert that the spec text contains the required rules — the strongest guarantee a
   prose rule can carry, and I won't dress it up as a behavioral test, since you were
   right that the previous ones only checked wording. If you'd rather have this enforced
   in the extractor, with tests that execute the decision (`book_to_skill/utils.py` owns
   the workdir, and its argument parser would take a flag), I'm glad to implement it
   there — that choice is yours to make, not mine to assume.

Gates on this head: `pytest -q` → 644 passed, 5 skipped (optional `html`/`epub` extras
not installed in my env), `ruff check .` → clean, `python3 tools/validate_skill.py
SKILL.md` → no Claude Code-breaking issues. The diff is two files, +49/−2: `SKILL.md`
and `tests/test_rerun_guard.py`.

virgiliojr94 (changes requested):
Destination resolution and missing-workdir fallback are fixed. One stale-input gap remains: replacing "Allow sharing." with "Avoid sharing." keeps the same filename and exact byte size, so the new reuse condition still accepts old extraction. metadata.json also stores file_size_mb rounded to two decimals, not an exact size. Require a recorded content fingerprint (plus source/config identity), or fall back to fresh extraction when unchanged content cannot be established. Keep the decision in the skill if appropriate; add a concrete changed-content recovery scenario rather than another wording assertion. Please refresh the PR description, which still quotes the removed unconditional-resume rule.

Scan-law4:
Thanks — round 2 is addressed on the branch, rebased onto current `master` (`c548330`, force-pushed).

**1. The stale-input gap you found is fixed at the root.** The extractor now records a per-source `sha256` (stdlib `hashlib`, streamed in 1 MiB chunks so memory stays constant on multi-GB inputs) and carries it into `metadata.json`'s `sources` list. Both extraction paths record it, including the pdf-inspector fast path — that path returns its own per-source dict for text-mode PDFs, so without a fingerprint there the gate would be blind for the most common input format. Hashing happens inside the same `try/except` that already wraps the stat call, so an unreadable file stays a per-source `ExtractionError` rather than aborting a batch.

**Your point (b) is what led us to drop size matching altogether.** Because `metadata.json` records `file_size_mb` rounded to two decimals, an exact byte-size comparison is not implementable — a ~12 KB file records as `0.01` MB ≈ 10,486 bytes, which never equals the real count. Rather than approximate it, the match is now filename + content fingerprint only: size is not part of the condition at all, and *any* content difference fails it, same-size or not.

**2. Concrete, executed recovery scenarios.** The decision itself is now a testable function — `reuse_is_safe()` in `book_to_skill/utils.py` — and `tests/test_fingerprint.py` executes every clause of its contract:

- `test_scenario_reuse_same_content` — unchanged content → reuse is allowed
- `test_scenario_changed_content_same_size_falls_back` — your exact case: `"Allow sharing."` → `"Avoid sharing."`, same filename, same byte size → `content changed for doc.txt`
- `test_scenario_sources_changed_falls_back` — a renamed input, and a changed input count
- `test_scenario_missing_workdir_falls_back` — temp cleanup after the run
- `test_scenario_mode_mismatch_falls_back` — recorded mode ≠ this run's mode
- `test_scenario_legacy_metadata_without_sha256_falls_back` — metadata predating fingerprints cannot establish freshness, so it starts a fresh extraction (the guard wording names this case explicitly)

Plus `test_sha256_file_streams_across_chunks` (>2 MiB, guarding the streaming loop against ever producing a *prefix* fingerprint — a truncated hash would be the same silent-stale-accept bug in a new costume).

**3. The decision stays in the skill**, as you suggested. Step 0 keeps it and now states the fingerprint standard, naming `reuse_is_safe()` as the executable form of that check.

**A test-architecture note.** While writing the hook-path regression test, our first draft stubbed `inspect_pdf` with a bare module-attribute assignment. It leaked into subsequent tests and produced six failures that appeared *only* in the full-suite run — your existing pdf-inspector suite caught it. The fix was to stay inside the architecture those tests already establish: `monkeypatch.setattr` for the stub, `_reset_state_for_tests()` for module state, and a fake-utils `SimpleNamespace` called through the installed hook. No new conventions and no refactoring of your scaffolding — the mistakes only surfaced because that scaffolding is strict enough to notice.

**Rebase.** While rebasing we found that `f08d343` (#228) had already added the same post-extraction stat-error `try/except` we had written independently, which made our copy redundant *and* conflicting. We kept yours and dropped ours, so this revision's contribution is purely additive: the fingerprint, the helper, the metadata field, the tests, and the guard wording.

Gates on `c548330`: `pytest -q` → **695 passed, 5 skipped** (current `master`: 675), `ruff check .` clean, `python3 tools/validate_skill.py SKILL.md` → no breaking issues. PR description refreshed.

virgiliojr94 (changes requested):
The same-size edit is now detected, but two reproduced paths still accept unusable/stale extraction: (1) reuse_is_safe returns True for an empty existing workdir with matching metadata supplied; it never checks full_text.txt or required extraction artifacts. (2) The hash is recorded only after parsing. Rewriting a source from "Allow sharing." to "Avoid sharing." after text is read produces old extracted text with the new file hash, and reuse_is_safe accepts it. Validate required artifacts and bind the fingerprint to the bytes actually extracted (a stable snapshot, or before/after validation that rejects changes), including the inspector path. Add regressions for both; the current happy-path fixture itself uses an empty workdir.

PR #225, opened 2026-09-13
fix(evals): stop scoring crashing on, and inventing counts from, recorded data

Hotragn:
## Summary

`tools/evals/score.py` calls `opens.index(target)` unguarded, so a trajectory whose `route_correct` was **recorded** by the harness while `opens` does not contain the target verbatim raises `ValueError` — and because `score()` maps over every trajectory, one such row aborts the entire scoring run. Separately, `isinstance(value, int)` accepts `True`, so a JSON `true` in a usage field is summed as a count.

## Type of change

- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does

- **Fixes (1):** an unguarded `list.index()` on a value that may legitimately be absent.
- **Fixes (2):** `bool` passing an `int` type check, which fabricates a usage number the module explicitly promises never to estimate.

## Motivation & context

Closes n/a — found while auditing the new `tools/evals` modules; no existing issue.

Both defects contradict the module's own stated contract. Its docstring says it scores *"without loading files or deriving missing observations"*, and `aggregate` says it *"never estimate[s] missing usage"*. The whole design — `_state()` keeping only explicit booleans, `UNKNOWN` for everything else — is about trusting what the harness recorded and refusing to infer. So:

- **(1)** is a cross-check the design says should not happen. `route_correct` is only *derived* from `opens` when the harness did not record it (`if "route_correct" not in observed`). When the harness *does* record it, the code proceeds to index `opens` by the target anyway. A harness that judges routing by its own criterion, or that records paths with any normalisation difference, is well within contract and gets a crash.
- **(2)** silently invents data. `bool` is a subclass of `int` in Python, so a recorded `true` is accepted as an observation, counted in `usage_observations`, and added to `recorded_usage` as 1.

The blast radius of (1) is what makes it worth fixing rather than tolerating: `score()` is a list comprehension over every trajectory, so a single unscoreable question takes down the whole report.

## How it works

**(1)** The position is computed once, guarded by membership:

```python
target_position = (
    opens.index(target)
    if isinstance(opens, list) and target is not None and target in opens
    else None
)
...
elif target_position is not None and target_position > 0:
    classification = "irrelevant_opens_before_target"
```

Absence now means what it should: there is no evidence of irrelevant opens *before* the target, so the classification falls through to `correct` — which is the right answer when all three recorded booleans are true. It does not override or second-guess the harness.

**(2)** A `_count()` helper mirroring the existing `_state()`:

```python
def _count(value: Any) -> Any:
    return value if isinstance(value, int) and not isinstance(value, bool) else UNKNOWN
```

Rejected alternative for (1): treating a missing target as `wrong_routing`. That would override an explicit recorded observation with an inference — the exact thing `_state()` exists to prevent — and would silently rewrite a harness's own judgement.

## Evidence

**Before / after**, `score_trajectory` with `route_correct` / `evidence_reached` / `answer_correct` all recorded true:

```
                                       before                     after
target not among opens                 CRASH ValueError           correct
empty opens list                       CRASH ValueError           correct
path normalisation differs ("./")      CRASH ValueError           correct
case differs                           CRASH ValueError           correct
target IS among opens (control)        correct                    correct
derived routing, target absent         wrong_routing              wrong_routing
```

**Usage counts:**

```
usage = {"calls": true, "input_tokens": 5, "output_tokens": 1}
  before -> {'input_tokens': 5, 'output_tokens': 1, 'calls': True}
  after  -> {'input_tokens': 5, 'output_tokens': 1, 'calls': 'unknown'}
```

and in `aggregate`, a recorded true alongside a genuine 3 previously produced `recorded_usage.calls == 4` from two "observations"; it now produces 3 from one.

**Tests** — new `tests/evals/test_score_robustness.py`, 19 cases in three groups:

- `TestTargetMissingFromOpens` — four parametrized absence cases, recorded flags passed through unchanged, and a three-question run proving one bad row no longer kills the other two.
- `TestDerivedRoutingUnchanged` — the regression net: target absent is still `wrong_routing`, target-first is `correct`, target-after-others is still `irrelevant_opens_before_target`, duplicate opens use the first position, missing answer is `unknown`, wrong answer still reported.
- `TestUsageCountsRejectBooleans` — true/false rejected per field, not summed into totals, genuine integers (including 0) still recorded, strings and floats still `unknown`.

RED against `master` (`01f8a74`): **11 failed, 8 passed**. GREEN:

```
$ python -m pytest tests/evals -q
49 passed in 3.75s

$ python -m pytest tests/ -q --ignore=tests/test_hermes_host_support.py
629 passed, 11 skipped in 19.27s

$ ruff check .
All checks passed!
```

## Screenshots / output

The before/after tables above are the output; this is a scoring function with nothing visual to capture.

## Checklist

- [x] `ruff check .` is clean

## Notes for reviewers

**On the `--ignore` in the test command above.** `tests/test_hermes_host_support.py` has **12 failures on `master` before this branch** — I verified that by stashing my change and re-running. They are a Windows artifact, not a product bug and not mine: those tests shell out to `bash -c` with Windows paths in `HOME` / `HERMES_HOME`, the candidate loop resolves nothing, and stdout comes back empty. They presumably pass on the Linux CI matrix. Flagging it only because it makes the suite red for anyone developing on Windows — happy to open a separate issue or send a portability fix if that is useful.

**Two smaller things in the same subsystem I left alone**, to keep this to one change — say the word and I will send either:

- `replay.py` does `from score import score`, a bare top-level import that only resolves when `tools/evals` itself is on `sys.path`, while the tests import `evals.score`. The two entry points disagree about the package layout.
- `replay()` reads `fixture["fixture_version"]` but `load_fixture()` validates only `schema_version` and `trajectories`, so a fixture missing `fixture_version` raises `KeyError` instead of the module's own `ValueError("unsupported fixture …")` style.

### DIFF
--- a/tests/evals/test_score_robustness.py
+++ b/tests/evals/test_score_robustness.py
@@ -1,0 +1,185 @@
+"""Scoring must consume whatever a harness recorded, without crashing or inventing.
+
+`tools/evals/score.py` documents itself as scoring "without loading files or
+deriving missing observations", and `aggregate` promises to "never estimate
+missing usage". Two things broke that contract.
+
+1. `opens.index(target)` was called unguarded. It is only reached when
+   `route_correct` and `answer_correct` are both true — but `route_correct` is
+   only *derived* from `opens` when the harness did not record it. A harness that
+   records `route_correct` itself, while `opens` does not contain the target
+   verbatim, raised `ValueError` and killed the entire scoring run rather than
+   one question.
+
+2. `isinstance(value, int)` accepted `True`, because `bool` subclasses `int` in
+   Python. A JSON `true` in a usage field was therefore treated as a recorded
+   count and summed as 1 — precisely the estimate the module promises not to make.
+"""
+
+import sys
+from pathlib import Path
+
+import pytest
+
+ROOT_DIR = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(ROOT_DIR / "tools"))
+
+from evals.score import UNKNOWN, aggregate, score, score_trajectory
+
+RECORDED = {"route_correct": True, "evidence_reached": True, "answer_correct": True}
+
+
+def _trajectory(question_id, target, opens, **observed):
+    return {
+        "question_id": question_id,
+        "expected": {"target": target},
+        "observed": dict(observed, opens=opens),
+    }
+
+
+class TestTargetMissingFromOpens:
+    """Recorded booleans must be trusted, not cross-checked into a crash."""
+
+    @pytest.mark.parametrize(
+        "opens, label",
+        [
+            (["chapters/ch01.md"], "a different file was opened"),
+            ([], "nothing was opened"),
+            (["./chapters/ch02.md"], "path normalisation differs"),
+            (["chapters/CH02.MD"], "case differs"),
+        ],
+    )
+    def test_does_not_crash(self, opens, label):
+        result = score_trajectory(
+            _trajectory("q", "chapters/ch02.md", opens, **RECORDED)
+        )
+
+        assert result["classification"] == "correct", label
+
+    def test_recorded_flags_are_passed_through_unchanged(self):
+        result = score_trajectory(
+            _trajectory("q", "b.md", ["a.md"], **RECORDED)
+        )
+
+        assert result["routing_correct"] is True
+        assert result["evidence_reached"] is True
+        assert result["answer_correct"] is True
+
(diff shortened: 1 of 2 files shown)


virgiliojr94 (approved):
Review findings addressed; validated against current master `2057f07`. Full pytest, Ruff and skill validation pass on the integration tree. The new regression fails on the baseline and passes here.

PR #228, opened 2026-09-17
fix(extractor): translate post-extraction stat errors

### LINKED ISSUE #227: [bug] Post-extraction file-size errors can still abort a batch
## What happened

`extract_single_file()` documents an important batch contract: every per-source
failure must surface as `ExtractionError`, because `main()` catches only that
exception before continuing with the remaining sources. PR #120 enforced this
contract for the initial magic-byte read, but the later file-size lookup still
calls `os.path.getsize()` without translating `OSError`.

If a source is removed, replaced, or becomes inaccessible after its content was
read but before metadata is assembled, `PermissionError`/`OSError` escapes with
a traceback and aborts the entire batch. The trusted pdf-inspector Markdown path
has the same unprotected metadata lookup.

## What you expected

A post-extraction file-size failure should be reported as `ExtractionError` for
that source. The batch runner should warn, skip it, and continue processing the
remaining inputs, consistently across the legacy and pdf-inspector paths.

## Source document
- Format: Markdown and PDF (synthetic fixtures)
- Pages / size: minimal fixtures; content is not relevant
- Language: English
- Does `python3 scripts/extract.py --check` show the relevant extractor installed? the PDF regression stubs the optional inspector result

## Repro

Run `extract_single_file()` on a readable Markdown file while making
`os.path.getsize()` raise `PermissionError("file became unavailable")`. This
deterministically models a file that changes between content extraction and the
final metadata lookup.

On `master` (`abc666b`), the raw exception escapes:

```text
book_to_skill/utils.py:1018: in extract_single_file
    file_size_mb = os.path.getsize(input_str) / (1024 * 1024)
PermissionError: file became unavailable
```

In a two-source `main()` run, the second valid source is never processed. The
same result occurs at `pdf_inspector_integration.py:146` when trusted native
Markdown takes the fast path.

## Environment
- OS: Windows 11 for reproduction; the `OSError` contract is platform-independent
- Python version: 3.12.10
- book-to-skill version / commit: `abc666b`

Stamina9:
## Summary (Required)

Translate file-size lookup failures into `ExtractionError` after content has
been extracted. A source that disappears or becomes inaccessible during a run
is now skipped without aborting the rest of the batch, including when the
pdf-inspector fast path is active.

## Type of change (Required)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Fixes:** The final `os.path.getsize()` calls in the legacy extractor and pdf-inspector result builder could leak `OSError` past the batch loop's `ExtractionError` boundary.
- **Adds:** Deterministic direct, batch-level, and pdf-inspector regressions for a source that becomes unavailable after content extraction.

The architecture promises that one bad source is skipped rather than ending a
multi-source run. PR #120 protected the initial format-sniff read, but a file can
still change after a successful read; successful content extraction does not
guarantee the later metadata stat will succeed. Both current extraction paths
need to uphold the same per-source exception contract.

Closes #227

Each post-extraction file-size lookup now catches `OSError` and raises a chained
`ExtractionError` naming the source. The existing batch loop handles that error
without any broader exception catch. Catching only `OSError` keeps programming
errors visible.

The change does not alter extraction selection, extracted text, successful
metadata values, or the batch error policy.

## Evidence (Required)
- **Tests:** Added a direct legacy-path translation test, a two-source `main()` test proving the good source still reaches `full_text.txt`, and a pdf-inspector fast-path translation test.
- **Before / after:** All three regressions fail on `master` with raw `PermissionError`; all pass after both metadata lookups enforce the existing exception contract.

```text
# Before
3 failed, 12 passed, 2 skipped in 0.08s

# After (targeted)
15 passed, 2 skipped in 0.05s

# Full clean Linux gate
662 passed, 8 skipped in 1.21s
```

```text
# Before: the raw exception terminates main()
PermissionError: file became unavailable

# After: the second source completes
WARNING: Skipping a.md: Could not read file size for a.md: file became unavailable
Extraction complete:
   Sources : 1 processed
```

- [x] `ruff check .` is clean

None. Successful size lookups return the same value. Only `OSError` from the
metadata lookup changes from an uncaught traceback to the documented per-source
failure type.

This is a follow-up to the same batch-resilience invariant addressed by #120,
but covers a distinct later operation and the newer pdf-inspector fast path.


virgiliojr94 (approved):
Validated the focused fix; validated against current master `abc666b`. Full pytest, Ruff and skill validation pass on the integration tree. The new regression fails on the baseline and passes here.

PR #229, opened 2026-09-17
fix(extractor): stop the Calibre path reusing a previous source's output

blrain3:
## Summary (Required)

`extract_with_ebook_convert()` wrote its conversion result to a fixed filename in the run's work directory and treated "exit code 0 and the file exists" as success. Because the work directory is per-process and shared by every source in a batch, a conversion that reported success without writing anything was satisfied by a file an earlier source had left behind — so one book's text was returned for a different source and recorded under *that* source's name. Each call now writes to its own filename, so the existence check can only be true for a file this call produced.

## Type of change (Required)

## What it does (Required)
- **Fixes:** a silent wrong-source extraction in the Calibre path. The root cause is that success was inferred from the *presence* of a fixed-name file rather than from *this invocation producing it*, and `OUTPUT_DIR` (`book_skill_work-<pid>`) is shared by all sources in one run. `parsers/calibre.py` never removed the artifact, so any later source in the same batch could read it. `metadata.json` still named the correct `source_file` / `extraction_method`, so nothing downstream could detect the mismatch.

Closes n/a — no issue open for this path. Filed as a PR because the fix is small and the evidence is reproducible; happy to split it into an issue first if you would rather triage it.

This is the same failure mode `test_per_run_workdir.py` describes for the shared work directory ("an agent ... could pick up a *different document's* extraction and build a skill from the wrong source"), reached through a different door: that fix made the *directory* per-run, but one file inside it still had a fixed name shared across sources.

It is also the same defect class as #219 (fixed in #220): success inferred from stale state left by an earlier attempt. #220 discards the in-memory entry before each attempt; this one cannot discard, because the stale state is a file on disk — so the fix is to stop giving two attempts the same destination. Flagging the difference in case you would prefer the discard-before-attempt shape here too (see the alternative below).

`book_to_skill/parsers/calibre.py` is also the one parser module with no test file, so this path had no coverage at all.

```python
output_path = OUTPUT_DIR / f"ebook-convert-output-{os.getpid()}-{next(_OUTPUT_SEQ)}.txt"
```

Each call gets a filename of its own, so `output_path.exists()` can only be true for a file this call wrote. The pid keeps it unique across runs sharing an explicit `BOOK_SKILL_WORKDIR`; the module-level `itertools.count()` covers repeated calls within one process.

Alternative I did not take: `unlink()` the artifact before converting. It closes the same window within a single process but reintroduces a fixed shared name across processes, so two runs pointed at the same `BOOK_SKILL_WORKDIR` could still interfere. **Happy to switch to it if you prefer that shape** — it is the smaller diff.

## Evidence (Required)

- **Tests:** new `tests/test_calibre_stale_output.py`, 5 tests, no Calibre install required (`shutil.which` / `subprocess.run` patched). Two of them fail on the pre-fix code and pass after it, which is what makes them regression tests rather than description:
  - `test_output_name_is_unique_per_call` — two conversions in one process must not share an output file.
  - `test_success_without_output_does_not_reuse_an_earlier_result` — a later conversion reporting success with nothing written must not receive the earlier source's text.
  - plus: a genuinely-written output is still returned; a non-zero exit is still `None`; a missing `ebook-convert` is still `None`.

- **Before / after:** the repro drives the real code path twice in one work directory — source A succeeds and writes; source B reports success and writes nothing.

```
# BEFORE (HEAD~1)
step 1 (bookA.mobi): returned 'TEXT OF BOOK A'
step 2 (bookB.mobi): ebook-convert exits 0 without writing output
  returned: 'TEXT OF BOOK A\n'
DEFECT REPRODUCED — book B was handed book A's text.

$ pytest tests/test_calibre_stale_output.py -q
2 failed, 3 passed          # the two assertions above

# AFTER
step 1 (bookA.mobi): returned 'TEXT OF BOOK A'
step 2 (bookB.mobi): ebook-convert exits 0 without writing output
  returned: None
NOT REPRODUCED — the stale artifact was not accepted for book B.

$ pytest tests/test_calibre_stale_output.py -q
5 passed

$ ruff check .                 # All checks passed!
$ pytest -q                    # 667 passed, 5 skipped
```

Also checked locally, since CI runs them: `ruff check --select E9,F --target-version py310 book_to_skill/ scripts/ tests/ tools/` (clean), `bandit -q -r book_to_skill scripts tools --severity-level high --confidence-level medium` (exit 0, same findings as before the change), the `smoke` job's exact commands (pass), and `ast.parse(..., feature_version=(3,9))` on the changed file for the 3.9 matrix leg.

Not applicable — internal behavior, no user-visible output change. The terminal before/after above is the equivalent.

- [x] `ruff check .` is clean

None for callers: the function's signature, return type, and `None`-on-failure contract are unchanged.

One thing to flag rather than decide — **the intermediate file's lifetime.** Previously a batch of N `.mobi` files left one `ebook-convert-output.txt` (each conversion overwrote the last); now it leaves N uniquely-named files. All of them live in the per-run temp work directory, which nothing currently cleans up, so this is tidiness rather than correctness — but if you would rather the intermediate not outlive the call at all, deleting it in a `finally` (or converting into a `TemporaryDirectory`) is a small follow-up. I have left the cleanup question out of this PR to keep the diff to the fix.

- The two design options above (unique name vs. delete-before-run) and the cleanup question are yours to call; the diff is written so switching is a small change.
- `parsers/calibre.py` remains the only parser without coverage beyond the new file — the paths I did *not* cover are the `except Exception` warn branch and `timeout=300` expiry. Say the word if you want those added here.


virgiliojr94 (changes requested):
The new tests pass, but PID + a process-local counter is not unique across runs. In two fresh interpreters sharing BOOK_SKILL_WORKDIR, with getpid() mocked to the same reused PID (4242), the second conversion exits 0 without writing and still returns the first process's text. Please allocate a fresh per-call TemporaryDirectory inside OUTPUT_DIR, read its output before cleanup, and add a reused-workdir regression. This also avoids leaving one full intermediate text per source.

blrain3:
Follow-up in 120b79d: the temp-dir creation failure now raises `ExtractionError` instead of a bare `OSError`. A raw `OSError` would bypass the per-source tolerance path in `utils.py` (which only catches `ExtractionError`, documented as non-fatal in batch mode) and could abort a whole batch instead of skipping one file. The original `OSError` stays chained via `raise ... from e`; the regression test asserts the new type. No change to the success path.

blrain3:
Addressed in `0fd041f` and `120b79d`: the Calibre path now converts into a per-call `TemporaryDirectory` under `OUTPUT_DIR` and reads before cleanup; `test_reused_workdir_with_same_pid_across_interpreters` reproduces your exact scenario (shared workdir, `getpid` mocked to 4242, module reloaded) and asserts the second conversion returns `None`. 14/14 checks green (py3.9–3.13 + ruff + bandit + CodeQL).

virgiliojr94 (changes requested):
Rechecked 120b79d: per-call isolation fixes stale output, and creation failures now become ExtractionError. Cleanup still escapes the error boundary: when TemporaryDirectory.__exit__ raises PermissionError after a successful conversion, extract_single_file propagates it and the batch (which catches only ExtractionError) aborts. Cover the entire temporary-directory lifecycle with the error boundary and add a cleanup-failure/batch-continuation regression. Also refresh the description: it still documents PID/counter names, retained intermediate files, and an unchanged None-on-failure contract.

PR #231, opened 2026-09-18
fix(epub): resolve encoded manifest hrefs

### LINKED ISSUE #230: [bug] Stdlib EPUB fallback skips encoded manifest hrefs
## What happened

The stdlib EPUB fallback reads each OPF manifest `href` as though it were the
literal ZIP member name. OPF hrefs are references, so a fragment such as
`chapter.xhtml#section-2` is not part of the archive filename, and percent
escapes such as `Chapter%201.xhtml` represent characters in that filename.

Both valid forms currently make `ZipFile.read()` look for a non-existent member.
The per-document exception handler then silently skips the chapter; if every
spine item uses one of these forms, extraction returns no text at all.

## What you expected

The dependency-free EPUB fallback should resolve manifest references to their
ZIP member names before reading them: decode XML entities and percent escapes,
discard query/fragment components, and still resolve the resulting path relative
to the OPF directory.

## Source document
- Format: EPUB (synthetic minimal fixtures)
- Pages / size: one spine item per fixture
- Language: English
- Does `python3 scripts/extract.py --check` show the relevant extractor installed? not applicable; this is the stdlib fallback used without ebooklib

## Repro

Create a minimal EPUB whose manifest contains either:

```xml
<item id="c1" href="chapter.xhtml#section-2" media-type="application/xhtml+xml"/>
```

or an OPF in `OEBPS/content.opf` with:

```xml
<item id="c1" href="Text/Chapter%201.xhtml" media-type="application/xhtml+xml"/>
```

while the corresponding ZIP members are `chapter.xhtml` and
`OEBPS/Text/Chapter 1.xhtml`. On `master` (`f08d343`), both focused regressions
return `None` instead of the chapter text.

## Environment
- OS: Windows 11 host / Linux clean-test environment
- Python version: 3.12.10
- book-to-skill version / commit: `f08d343`

Stamina9:
## Summary (Required)

Resolve OPF manifest hrefs before the stdlib EPUB fallback reads ZIP members.
Chapters referenced with fragments or percent-encoded filenames are no longer
silently omitted.

## Type of change (Required)
- [ ] 📝 Docs only
- [ ] 🔒 Security
- [ ] 🧹 Chore / CI / tooling

## What it does (Required)
- **Fixes:** The stdlib EPUB fallback treated OPF manifest references as literal ZIP member names, dropping content when an href contained a fragment or percent escape.
- **Adds:** Synthetic regressions for a fragment-bearing href and a percent-encoded filename under an OPF subdirectory.

The OPF manifest describes references, while `ZipFile.read()` requires the exact
archive member name. Passing the raw href across that boundary makes otherwise
readable chapters disappear without a warning because the existing per-item
fallback intentionally skips unreadable members.

Closes #230

A small resolver now mirrors the normalization a parsed OPF reference needs:

1. decode XML character/entity references;
2. discard query and fragment components;
3. percent-decode the path; and
4. resolve it relative to the OPF directory with POSIX ZIP-path semantics.

The normalized member name is used consistently for spine ordering, de-duplication,
the manifest safety net, and archive reads. EPUBs with plain relative hrefs keep
the same paths and behavior. No runtime dependency is added.

## Evidence (Required)
- **Tests:** Added two synthetic one-chapter EPUB regressions; no copyrighted text or binary fixture is committed.
- **Before / after:** Both focused tests fail on current `master` because extraction returns `None`; both pass after reference normalization.

```text
# Before on master f08d343
2 failed in 0.13s

# Full clean Linux gate after the fix
664 passed, 8 skipped in 1.09s
```

```text
# Before
extract_with_zipfile(...) -> None

# After
extract_with_zipfile(...) -> chapter text
```

- [x] `ruff check .` is clean

None. Plain relative hrefs resolve identically. The change only makes standards-style
reference syntax reach the archive member it already names.

The implementation stays in the dependency-free fallback and uses only Python's
standard library.


virgiliojr94 (approved):
Validated against current master `91d17c2`: full pytest, Ruff, skill validation and synthetic CLI smoke pass on the integration tree. The new regressions fail on the baseline and pass here.

